PR #26, opened 2015-06-12
83/h/story 2550 Add filtering functionality

colin-allen:
Added filtering implementation and tests.
@sherriman 
@InvokeJLa 
@codyjca 

Would you please do a code review!  Thanks, Colin.

### DIFF
--- a/dist/backdraft.js
+++ b/dist/backdraft.js
@@ -966,7 +966,7 @@ _.extend(Plugin.factory, {
       // returns all models matching the current filter criteria, regardless of pagination
       // since we are using deferred rendering, the dataTable.$ and dataTable._ methods don't return all
       // matching data since some of the rows may not have been rendered yet.
-      // here we use the the aiDisplay property to get indecies of the data matching the currenting filtering
+      // here we use the the aiDisplay property to get indicies of the data matching the current filtering
       // and return the associated models
       return _.map(this.dataTable.fnSettings().aiDisplay, function(index) {
         return this.collection.at(index);
@@ -1016,6 +1016,7 @@ _.extend(Plugin.factory, {
     _dataTableCreate : function() {
       this.dataTable = this.$("table").dataTable(this._dataTableConfig());
       this._installSortInterceptors();
+      this._createFilterControls();
       this.reorderableColumns && this._enableReorderableColumns();
       this._columnManager.on("change:visibility", this._onColumnVisibilityChange);
       this._columnManager.applyVisibilityPreferences()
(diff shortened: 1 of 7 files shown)


src/plugins/data_table/server_side_data_table.js
```diff
@@ -90,6 +90,12 @@ var ServerSideDataTable = (function() {
 
     _fetchServerData : function(sUrl, aoData, fnCallback, oSettings) {
       var self = this;
+      if (this.serverSideFiltering) {
+        var filterJson = {};
+        filterJson.name = "ext_filter_json";
+        filterJson.value = this._getFilteringSettings();
+        aoData.push(filterJson);
```
InvokeJLa: Can we turn this into a single line, since filterJson is never used again?

spec/plugins/data_table/server_side.js
```diff
@@ -433,4 +434,57 @@ describe("DataTable Plugin", function() {
...
   });
 
+  describe("filtering", function() {
+    beforeEach(function() {
+      table = new app.Views.T({ collection : collection });
+      table.render();
+      jasmine.Ajax.requests.mostRecent().response(mockResponse.get());
+    });
+
+    it("should have filtering info for filterable column in column manager", function() {
```
InvokeJLa: "should contain an object that describes the type of filter to be applied and possible values if applicable." or something more descriptive.

spec/plugins/data_table/server_side.js
```diff
@@ -433,4 +434,57 @@ describe("DataTable Plugin", function() {
...
+        if (col) {
+          numCols++;
+          if (col.filter)
+            numFilters++;
+        }
+      }
+      //console.log("Number of columns found in config: "+numCols);
+      //console.log("Number of columns with filters:  "+numFilters);
+      expect(numCols).toEqual(4);
+      expect(numFilters).toEqual(2);
```
InvokeJLa: We should also test to make sure the types are actually getting put into the ColumnManager filter object/settings. I'd also include some types that aren't free form input text.

spec/plugins/data_table/server_side.js
```diff
@@ -433,4 +434,57 @@ describe("DataTable Plugin", function() {
...
+      //var cg = table._columnManager.columnsConfig();
+      table.dataTable.find("thead th").each(function (index) {
+        var title = this.outerText;
+        title = title.substr(0, title.indexOf('Filter'));
+        var col = cg.columnConfigByTitle.attributes[title];
+        if (col && col.filter) {
+          if (col.filter.type == "string") {
+            $('#menu-name input', this).val("Scott").trigger("change");
+            expect(col.filter.value).toEqual("Scott");
+            expect(jasmine.Ajax.requests.mostRecent().url).toMatch("ext_filter_json");
```
InvokeJLa: Cool. We would probably want verify the values of the filter, not only if it exists.

src/plugins/data_table/data_table.js
```diff
@@ -12,7 +12,8 @@ var LocalDataTable = (function() {
       this.options = options || {};
       // copy over certain properties from options to the table itself
       _.extend(this, _.pick(this.options, [ "selectedIds" ]));
-      _.bindAll(this, "_onRowCreated", "_onBulkHeaderClick", "_onBulkRowClick", "_bulkCheckboxAdjust", "_onDraw", "_onColumnVisibilityChange");
+      _.bindAll(this, "_onRowCreated", "_onBulkHeaderClick", "_onBulkRowClick", "_bulkCheckboxAdjust", "_onDraw",
+          "_onColumnVisibilityChange", "_onReorder");
```
InvokeJLa: Is '_onReorder' too generic of a name? Colin said there is a separate function in the reordering plugin with a similar name. We came with '_onBdColumnReorder' as a possible other function name.

src/plugins/data_table/data_table.js
```diff
@@ -261,24 +269,181 @@ var LocalDataTable = (function() {
...
+    //   equal to.  these text inputs need to be uniquely identified so we can filter on
+    //   values entered into more than one of them
+    // * list filtering requires a list of labeled checkboxes for each filter option
+    //   these checkboxes need to be identified by the value they represent
+    // The IDs "value", "gt", "lt" and "eq" are used to determine in what element in the
+    // filter object in the column manager we store the value entered by the user
+    _generateFilteringControls: function(head, col) {
+      var filter = col.filter;
+      if (filter.type == "string") {
+        $(head).append('<input class="filter-string" id ="value" type="text" placeholder="Search ' +
```
InvokeJLa: For these lines, can we put the raw HTML into JS templates? And then use Underscore's templating lib to interpolate?

dist/backdraft.js
```diff
@@ -826,11 +826,30 @@ _.extend(Plugin.factory, {
       <table cellpadding="0" cellspacing="0" border="0" class="table table-striped table-bordered"></table>\
```
colin-allen: You can ignore this file in the context of this pull request, it's a `grunt build` of the other files in the repository.

colin-allen:
@nburwell Scott told me to tag you in on this.  I know you're on the road but if you get the chance feel free to give this a look.  :cactus:

dist/backdraft.js
```diff
@@ -1079,24 +1109,177 @@ _.extend(Plugin.factory, {
...
+    //   equal to.  these text inputs need to be uniquely identified so we can filter on
+    //   values entered into more than one of them
+    // * list filtering requires a list of labeled checkboxes for each filter option
+    //   these checkboxes need to be identified by the value they represent
+    // The IDs "value", "gt", "lt" and "eq" are used to determine in what element in the
+    // filter object in the column manager we store the value entered by the user
+    _generateFilteringControls: function(head, col) {
+      var table = this;
+      var filter = col.filter;
+      if (filter.type == "string") {
```
codyjca: I believe we prefer === over == for comparisons in js.

dist/backdraft.js
```diff
@@ -1320,6 +1506,48 @@ _.extend(Plugin.factory, {
...
+
+    // gets an object representing all filtering settings set in the column
+    // manager to send to the backend to retrieve a filtered dataset
+    _getFilteringSettings: function() {
+      var table = this;
+      var result = [];
+      var cg = this._columnManager._configGenerator;
+      for (var i = 0; i < cg.columnsConfig.length; i++) {
+        var col = cg.columnsConfig[i];
+        if ((col.filter) ) {
```
codyjca: Extra parenths needed?

spec/plugins/data_table/server_side.js
```diff
@@ -433,4 +435,104 @@ describe("DataTable Plugin", function() {
...
+      var hasFilter = [];
+      table.dataTable.find("thead th").each(function (index) {
+        hasFilter.push(this.getElementsByClassName('DataTables_filter_wrapper').length > 0);
+      });
+      expect(hasFilter).toEqual([false, true, true, true, false]);
+    });
+
+    it("should track filtering in column manager and in ext_filter_json parameter", function() {
+      var cg = table.configGenerator();
+      //var cg = table._columnManager.columnsConfig();
```
codyjca: Still need this comment?

src/plugins/data_table/data_table.js
nburwell: this is down in the plugins, I think "Bd" is redundant / not needed.. but I like going with `_onColumnReorder`

dist/backdraft.js
nburwell: we do, good catch

spec/plugins/data_table/server_side.js
```diff
@@ -433,4 +435,104 @@ describe("DataTable Plugin", function() {
...
+      var expectedFilterObj = [];
+      table.dataTable.find("thead th").each(function (index) {
+        var title = this.outerText;
+        title = title.substr(0, title.indexOf('Filter'));
+        var col = cg.columnConfigByTitle.attributes[title];
+        if (col && col.filter) {
+          if (col.filter.type == "string") {
```
nburwell: here is the `==` that should be `===`

spec/plugins/data_table/server_side.js
```diff
@@ -433,4 +435,104 @@ describe("DataTable Plugin", function() {
...
+            VerifyFilterAjax(expectedFilterObj);
+
+            // test unassignment
+            $('.filterMenu input', this).val("").trigger("change");
+            expect(col.filter.value).toEqual(null);
+            // verify ajax
+            expectedFilterObj = [];
+            VerifyFilterAjax(expectedFilterObj);
+          }
+          else if (col.filter.type == "numeric") {
```
nburwell: and here.. `===` and below

src/plugins/data_table/data_table.js
```diff
@@ -261,24 +291,177 @@ var LocalDataTable = (function() {
...
+    //   equal to.  these text inputs need to be uniquely identified so we can filter on
+    //   values entered into more than one of them
+    // * list filtering requires a list of labeled checkboxes for each filter option
+    //   these checkboxes need to be identified by the value they represent
+    // The IDs "value", "gt", "lt" and "eq" are used to determine in what element in the
+    // filter object in the column manager we store the value entered by the user
+    _generateFilteringControls: function(head, col) {
+      var table = this;
+      var filter = col.filter;
+      if (filter.type == "string") {
```
nburwell: more `===` needed in this function

src/plugins/data_table/data_table.js
```diff
@@ -261,24 +291,177 @@ var LocalDataTable = (function() {
...
+        $(head).append(table.filterStringTemplate( { title: col.title } ));
+      } else if (filter.type == "numeric") {
+        $(head).append(table.filterNumericTemplate());
+      } else if (filter.type == "list") {
+        var checkList = '<ul>';
+        for (var i = 0; i < filter.options.length; i++)
```
nburwell: should we raise if filter.type === 'list' and no `options` key passed? otherwise this will be a undefined.length exception

dist/backdraft.js
colin-allen: Ok, will make sure to change all instances of == used for comparisons with ===

dist/backdraft.js
colin-allen: No, not sure why that's there!

spec/plugins/data_table/server_side.js
colin-allen: Nope, will clean it out

src/plugins/data_table/data_table.js
colin-allen: _onColumnReorder it will be!

src/plugins/data_table/data_table.js
colin-allen: That's a good idea.  Ideally it should never happen because it should throw an error on the backend, but I don't know that for certain and it's always good to be safe.  Will take care of that.

Do you want to throw an exception or do you want to just not make the filtering controls if filter.options is undefined when filter.type === 'list'?

src/backdraft.js
```diff
@@ -3,6 +3,13 @@
   var Backdraft = {};
   Backdraft.Utils = {};
 
+  // use squiggly braces for underscore templating so we don't conflict with ruby templating
+  _.templateSettings = {
+    evaluate    : /{{([\s\S]+?)}}/g,
+    interpolate : /{{=([\s\S]+?)}}/g,
+    escape      : /{{-([\s\S]+?)}}/g
+  }
+
```
colin-allen: Discussed with @nburwell 
This will have to do for now.  We can consider going back to the defaults in the future, but it will require a STORY on invoca/web/

src/plugins/data_table/data_table.js
nburwell: I think this is fine for now. Should we do the same when options.length === 0 ? No point creating a filter UI with an empty list..

src/backdraft.js
colin-allen: Added note to readme about using ``{{ }}`` templating instead of default ``<% %>`` templating.

nburwell:
This looks good :+1: 

Worth doing as a follow-on PR, that I forgot to mention/catch earlier:
We need support for setting initial states to these filters. I.e. when a page is reloaded from a URL, it may already have some filters and filter values set.  Please see what Euge did for supporting initial state in the ActionBar plugin (part of ActionTable), and may make sense to copy that pattern and naming.

PR #27, opened 2015-06-30
85 WEB-1457 save datatable user preferences

zmathe-sv:
@nburwell please take a look to see if this solution is going in the right direction.
It consists of the reportSettings being passed down to DataTable component. It then gets applied to columnsConfig and sorting.
```
reportSettings = {
  reportType: "manage_advertiser_campaigns",
  updateUrl: "/users/report_settings/6",
  columnAdds: ["col1_attr", "col2_attr"],
  columnSubtracts: ["col3_attr"],
  sorting: [ ["col1_attr", "desc"] ]
}
```
As for persisting the preferences, I'm planning to use the same DataTable component. There are events for column order, visibility change, onDraw (for sorting). When they trigger and there is a state change, a request to backend will be made using `reportSettings.updateUrl` and corresponding changed settings.

As of now, the state saving mechanism is controlled by the `reportSettings` options parameter. Didn't add a separate flag to turn it on, so whenever a DataTable is initialized with `reportSettings`, it will persist settings and apply them on initialize.

The other part of the solution is in the web repo: [web/pull/3383](https://github.com/Invoca/web/pull/3383)

CC: @DanBradbury


dist/backdraft.js
```diff
@@ -575,8 +581,101 @@ _.extend(Plugin.factory, {
         return columnType;
       }
     });
+  },
+
+  _applySavedState: function() {
+    var reportSettings = this.table.options.reportSettings;
```
InvokeJLa: While I understand that the user settings are being stored in the Invoca ReportSettings model/table, these changes are being made to the Backdraft public repo. Can we rename references in Backdraft to something more generic? Like "dataTableSettings?" or "userPreferences?" or something like that?

dist/backdraft.js
```diff
@@ -681,6 +781,86 @@ _.extend(Plugin.factory, {
...
+      url  : this.reportSettings.updateUrl,
+      method: 'POST',
+      data : _.extend({
+        ajax          : 1,
+        format        : 'json',
+        report_type   : this.reportSettings.reportType
+      }, settingsUpdate)
+    });
+    this._cacheSettingsState(currentSettings);
+  }
```
InvokeJLa: Do we want to make Backdraft always save back to the server, each time a user takes an action like a column reorder or a column visibility change? For Invoca's stuff, yes, but for Backdraft in general? If so, can we generalize the AJAX call so that the user can specify method, and parameters?

src/plugins/data_table/settings_manager.js
```diff
@@ -0,0 +1,79 @@
...
+        };
+        this._saveSettings(updates, currentSettings);
+      }, this);
+    }
+  },
+
+  _onColumnOrderChange: function() {
+    currentSettings = this._getCurrentSettingsState();
+    if (!_.isEqual(this.settingsState.columnOrder, currentSettings.columnOrder)) {
+      this._saveSettings({ column_order: currentSettings.columnOrder }, currentSettings);
```
InvokeJLa: Here, the call to _saveSettings will always fire if when a user reorder columns. Can we add a check like:

```
if (this.reportSettings.updateUrl)
      this._saveSettings(etc, etc)
```

dist/backdraft.js
nburwell: agreed this needs to be very general down in backdraft. We do not want anything coupled to our application logic here. Doe saveSettings always get called? if you dont have a reportSettings.updateUrl is it going to blow up and/or cause 404s?

At a minimum this all needs to be optional and parameterized (including extra data params like the ajax=1 and format:json and report_type).

But @InvokeJLa hadn't we gone with doing these settings saving things in action_table plugin, in the web repo? (for your report work?)

Wherever you did work to save the show/hide, I want it to be the same pinch point, not a second way of updating settings.

dist/backdraft.js
nburwell: agreed (if we keep this functionality down in backdraft at all).  I like `persistedDataTableSettings`..?

dist/backdraft.js
```diff
@@ -575,8 +581,101 @@ _.extend(Plugin.factory, {
...
+          column.visible = true;
+        }
+      }, this);
+
+      _.each(reportSettings.columnSubtracts, function(columnName) {
+        column = this._findColumnByAttr(columnName);
+        if (column) {
+          column.visible = false;
+        }
+      }, this);
```
InvokeJLa: This section seems to deal exclusively with column visibility. Can we break that out into a separate method? Then the contents of _applySavedState can become: 
```
_applySavedState: function() {
   ... snip... 
   this._setColumnVisibility(reportSettings);
   this._reorderColumns(reportSettings.columnOrder);
}
```

src/plugins/data_table/column_config_generator.js
```diff
@@ -86,5 +92,97 @@ var ColumnConfigGenerator =  Backdraft.Utils.Class.extend({
...
+          column.visible = true;
+        }
+      }, this);
+
+      _.each(reportSettings.columnSubtracts, function(columnName) {
+        column = this._findColumnByAttr(columnName);
+        if (column) {
+          column.visible = false;
+        }
+      }, this);
```
InvokeJLa: This section seems to deal exclusively with column visibility. Can we break that out into a separate method? Then the contents of _applySavedState can become: 
```
_applySavedState: function() {
   ... snip... 
   this._setColumnVisibility(reportSettings);
   this._reorderColumns(reportSettings.columnOrder);
}
```

src/plugins/data_table/settings_manager.js
```diff
@@ -0,0 +1,79 @@
+var SettingsManager = Backdraft.Utils.Class.extend({
```
InvokeJLa: I like having all user preference related code in one place, like the SettingsManager. However, this may become unnecessary if this behavior is integrated into ActionTable. The hooks into the events  ```table.on("draw"... ```, ```table._columnManager.on("etc"... ``` can be created inside of the ActionTable plugin.

dist/backdraft.js
InvokeJLa: @nburwell Yes, we did this in the action_table plugin for column visibility, and I agree we want a single entry for all writes back to the server.

I do however, like that there is a single object that handles all of the write backs to the user settings (``` SettingsManager```). I think that's something I'd like to import over to ActionTable.

zmathe-sv:
@InvokeJLa Followed your suggestions and removed settings manager now that's included in ActionTable. Here I added functionality to set columns order, sorting. These are used when restoring the persisted settings or when restoring the default ones.
Can you take another look? Thanks!

InvokeJLa:
Looks ok to me.

dist/backdraft.js
```diff
@@ -934,6 +950,28 @@ _.extend(Plugin.factory, {
...
 
+    columnOrder: function(order) {
+      if (this.reorderableColumns) {
+        this._changeColumnOrder(order);
+      }
+    },
+
+    restoreColumnOrder: function() {
+      if (this.reorderableColumns) {
+        this._changeColumnOrder({ reset: true});
```
nburwell: white space nit: missing whitespace after `true`

nburwell:
:+1: nice. just makes me want documentation on this even more!  (a project Euge and I were working on when he left. I do have some to share & extend on the plugin side in web repo.. but we're lacking it here on backdraft side)

PR #28, opened 2015-07-08
85 STORY-2678 Finish Backdraft Parity: Initial commit.

colin-allen:
Made new templates for handling filter action button, filter dropdown menu button.  Filter options button fixed to use correct templating.  Filtering now has a dropdown menu that activates on click and has Filter and Clear buttons.  Filtering now has a image button instead of the word filter, button changes when filter is active/inactive.  Filter menu appears using a slide animation.  CSV exporting now supported.

[Review on Reviewable](https://reviewable.io/reviews/invoca/backdraft/28)


colin-allen:
Need to look at Zsolt's stuff to see what patterns I can draw into this.  Also need to look into setting initial filter state as mentioned at the end of the discussion for #26 .

dist/backdraft.js
```diff
@@ -845,11 +845,19 @@ _.extend(Plugin.factory, {
     ',
     sFilterListTemplate : '\
       <li><label><input class="list" id="value" type="checkbox" name="{{= attr }}" \
-         value="{{= options }}" /> <%= options %></label></li> \
+         value="{{= options }}" /> {{= options }}</label></li> \
+    ',
+    sActionButtonTemplate : '\
+      <button class="btn btn-primary btn-sm {{= klass}}" name="button" type="submit" title="">{{= caption}}</button>\
```
colin-allen: called it klass because class causes compiling issues.  if you have a better idea for the name let me know.
also, ActionButton isn't the best name for this control.  any other thoughts on that would be appreciated.

dist/backdraft.js
```diff
@@ -845,11 +845,19 @@ _.extend(Plugin.factory, {
...
+    ',
+    sFilterButtonTemplate : '\
```
colin-allen: could be renamed?  I had trouble thinking of a better name, let me know what you think.

dist/backdraft.js
```diff
@@ -1153,19 +1163,20 @@ _.extend(Plugin.factory, {
     // The IDs "value", "gt", "lt" and "eq" are used to determine in what element in the
     // filter object in the column manager we store the value entered by the user
     _generateFilteringControls: function(head, col) {
-      var table = this;
```
colin-allen: this variable is unnecessary.

dist/backdraft.js
```diff
@@ -1178,31 +1189,63 @@ _.extend(Plugin.factory, {
...
+        var filterIcon = $('.DataTables_filter_wrapper span', head);
         if (filter.type === "list") {
           if (this.checked) {
             filter[this.id] = filter[this.id] || [];
             filter[this.id].push(this.value);
+            filterIcon.removeClass("filterInactive");
+            filterIcon.addClass("filterActive");
           }
-          else {
+          else if (filter[this.id]) {
```
colin-allen: necessary because of an edge case where filter[this.id] is not defined.

nburwell:
> Need to look at Zsolt's stuff to see what patterns I can draw into this. Also need to look into setting initial filter state as mentioned at the end of the discussion for #26 .

Sounds good. Zsolt's work supported setting initial state for column order, sort etc so I hope it can be in that same persistedDataTableSettings object that he was initializing and passing around. Maybe `persistedDataTableSettings.filters` ?

dist/backdraft.js
```diff
@@ -1214,42 +1257,67 @@ _.extend(Plugin.factory, {
...
+      $('button', head).appendTo(filterDiv);
+      wrapperDiv.appendChild(filterDiv);
       head.appendChild(wrapperDiv);
 
-      // handle hovering on wrapperDiv
-      $('.DataTables_filter_wrapper', head).hover(function () {
-        $('.filterMenu', head).slideDown(200);
-      }, function () {
-        $('.filterMenu', head).slideUp(50);
+      // handle filter menu display
```
colin-allen: this code isn't very pretty.  if you have any suggestions on refactoring let me know.  tried using the functionality of bootstrap's dropdown menus, but ran into issues with the menus closing whenever you clicked a control in them (impeding functionality).  might be a clever way to fix that using events, but none of the stack overflow solutions I found worked.

look into using bootstrap popover elements?

dist/backdraft.js
```diff
@@ -1476,6 +1544,70 @@ _.extend(Plugin.factory, {
...
+        filterJson.value = this._getFilteringSettings();
+        aoData.push(filterJson);
+      }
+      $.ajax({
+        url: sUrl,
+        data : aoData,
+        dataType : "text",
+        cache : false,
+        type : this.ajaxMethod || "GET",
+        success: function(response, status, xhr) {
```
colin-allen: this code comes from stack overflow.  it starts a download for the CSV we get back.

dist/backdraft.js
```diff
@@ -1214,42 +1257,67 @@ _.extend(Plugin.factory, {
     // @col: The column from the column manager corresponding to the column we're creating
     //   filtering wrappers for.
     _createFilteringWrappers: function(head, col) {
```
colin-allen: could we entirely remove the use of document.createElement by using some larger-scale templates?  (templates with templates in them)?

colin-allen:
Will look into persistedTableSettings.  Right now stuff is stored in the column manager.  Initial filter values are possible, albeit unleveraged/tested right now.

colin-allen:
In addition to Nick, I'd like to welcome @zmathe-sv and @InvokeJLa to code review this.

dist/backdraft.js
nburwell: I dont think we want ajax here and all this work. At least for now I'd rather remove all this code and mirror / use our existing reporitng CSV download infrasctucture, which is to do a window.location to the same url but with .csv extension, and server responds with the disposition / attachment headers and browser does the right thing.

Rather not use a different method here for now.

dist/backdraft.js
zmathe-sv: Wouldn't be better to create separate Views for the filters and encapsulate templates and events in there? There's something similar in ActionTable when setting up ActionBar filters.
As for the `window.activeMenu` variable, that seems to be too intrusive. I would store it in the table instead.

dist/backdraft.js
colin-allen: I'm pretty sure we _can't_ do that due to how we're passing parameters for how to make the table (in a post request).  Setting window.location is equivalent to doing a get request at a URL, which doesn't work for getting table data within the reporting infrastructure we're creating with DataTables/Backdraft.

If I could have just set window.location, I would have done that.

dist/backdraft.js
colin-allen: Good call, will look into doing that.

colin-allen:
@euge: The reasoning behind allowing a reset even if addData is not defined is so we can clear the collection manually.  This happens in the specific instance where we have a modal view on a page with a dataTable in it.  The issue I found was as follows:

- Modal view is opened, it instantiates a server side dataTable.
- Modal view is closed.
- Modal view is re-opened with a different data set, but fails because a server-side dataTable requires an empty collection (and our collection is not empty because of the previous request)

I got the expected behavior with my alteration to the reset function as shown here and calling collection.reset (with collection being the variable holding my collection).  Would it be better to make a separate method for this circumstance?

--

Regarding doAjaxUpdate, I think I missed that function.  Will try it and see if it works.  Thanks.

euge:
@colin-allen Gotcha! I think i see what you mean, but if the modal is closed, doesn't the dataTable within it get nuked? Is the collection shared across iterations of the modal? If it is, I would see if there's any way you can just create the necessary collection as part of creating the modal rather than being passed in. I think we really want to ensure that collections that are the basis for a server side table are not really interacted with directly by the user. If you look at the source in `server_side_data_table.js` you can see that "_onAdd` and `_onRemove` both raise exceptions in order to avoid letting users think they can modify it.

colin-allen:
@euge: Yes, the dataTable does get nuked, but the collection is shared.  So you're thinking we'd make the collection on a per-modal basis... That makes sense.  I'll give a try.

I can understand the reasoning behind not letting the user interact with the data in a server-side data table.  That makes a lot of sense.  I'll see if I can find a better way to handle this situation.

colin-allen:
@euge: Making the collection on a per-modal basis totally fixed the problem, thanks!  Seems the original design for what I was working with was at fault.

euge:
@colin-allen wooooot. Happy to help!

dist/backdraft.js
nburwell: we should discuss directly here.. we do this for the old tables.  The point is you send a URL over with all the correct sorting, filtering, etc and the server will respond with the CSV file.  There is nothing you are "getting out of the current table" other than what filters and option params are set.

colin-allen:
@nburwell can you please code review this and address my questions.

dist/backdraft.js
nburwell: I didn't catch this before but we don't do hungarian notation for variable names.. (dataTables does/did but I think their 1.10 got rid of it and for our interfaces we don't expose that).

Can you remove the leading `s` on these? 

What about a nested structure here? I.e.:

```
filterTemplates : {
  string : '...',
  numeric : '...',
  list : '...'
},
```

If the action button template for filters only? I think so far it is?

So could include in that nesting and clarify the other items as the input portion of the templates?

```
filterTemplates : {
  stringInput : '...',
  numericInput : '...',
  listInput : '...',
  actionButton : '...'
},
```

nburwell: Instead of using `klass` we often use `className` (and I am pretty sure backbone uses that too for when we need a variable name for a CSS class)

dist/backdraft.js
nburwell: is this the icon that toggles the filter UI?  I agree this name is too general. Assuming you nest it in the filterTemplates hash it could be toggleButton or something that describes what this item is doing

dist/backdraft.js
```diff
@@ -1178,31 +1190,63 @@ _.extend(Plugin.factory, {
...
+
+      // update columnManager filter when input changed
       $('input', head).on('change', function () {
+        var filterIcon = $('.DataTables_filter_wrapper span', head);
         if (filter.type === "list") {
           if (this.checked) {
             filter[this.id] = filter[this.id] || [];
             filter[this.id].push(this.value);
+            filterIcon.removeClass("filterInactive");
+            filterIcon.addClass("filterActive");
```
nburwell: Can you re-render the template rather than tweaking classes manually like this?  The point of a template is so you can change state (like active or not) and simply re-render

dist/backdraft.js
nburwell: it's not clear to me in this context what `this` is or what `this.id` is, and seems like too much inline code here with intricate if / else / else if cases.. hoping some of this logic can be extracted and encapsulated in the templates themselves (and maybe each filter type should be its own View).

Either way can you make it more clear in this code what is going on here and what this.id is? (maybe name it a temporary variable that is better named?

dist/backdraft.js
```diff
@@ -1178,31 +1190,63 @@ _.extend(Plugin.factory, {
...
           filter[this.id] = this.value;
+          filterIcon.removeClass("filterInactive");
+          filterIcon.addClass("filterActive");
         }
+      });
+
+      // update table when filter button clicked
+      $('.btn-filter', head).on('click', function() {
+        $("input", head).each(function() {
+          if (this.type === "text") {
```
nburwell: maybe I'm missing something but can this also be:

```
$("input[type=text]", head).each(function() {
  $(this).trigger("change");
}
```

and then it simply becomes:

```
$("input[type=text]", head).trigger("change");
```

dist/backdraft.js
nburwell: Why does "using vies" mean it has to be done outside of Backdraft repo?  Couldn't they be backdraft views, which are pretty similar to a backbone view? 

Because I agree with Zsolt's suggestion to have the templates and event handling within views, and much less code hand-written in the the DataTable class(es)

dist/backdraft.js
```diff
@@ -1214,42 +1258,68 @@ _.extend(Plugin.factory, {
...
-        $('.filterMenu', head).slideDown(200);
-      }, function () {
-        $('.filterMenu', head).slideUp(50);
+      // handle filter menu display
+      $(document).click(function (event) {
+        var className = event.target.classList[0];
+        if (event.target.offsetParent) {
+          var parentClassName = event.target.offsetParent.classList[0];
+        }
+        if ((table.activeMenu) && (((className !== 'filterMenu') && (parentClassName !== 'filterMenu')) || (className === 'btn'))) {
```
nburwell: this line is hard to read. which element(s) get this slide up behavior and can you just put an explicit class on them that you can read, without having to also look at a parent class??

dist/backdraft.js
```diff
@@ -1214,42 +1258,68 @@ _.extend(Plugin.factory, {
     // @col: The column from the column manager corresponding to the column we're creating
     //   filtering wrappers for.
     _createFilteringWrappers: function(head, col) {
+      var table = this;
+      table.activeMenu = null;
```
nburwell: better that its on the table now.. but how is it different from filterActive and can it be named activeFilterMenu at least?

dist/backdraft.js
```diff
@@ -1214,42 +1258,68 @@ _.extend(Plugin.factory, {
...
       var filter = col.filter;
+      var filterActive = filter.value || filter.eq || filter.gt || filter.lt;
```
nburwell: related to above.. is this the active filter, or just a boolean? I think the latter.. so how about isFilterActive ?

dist/backdraft.js
nburwell: I'd like to at least have you explore popover elements if you haven't already. By default they don't dismiss by clicking outside of them, or elements inside them (which I think is a good thing.. since you have an explicit cancel button)

dist/backdraft.js
nburwell: exportData feels more like an application concern at this point (ie maybe should live in actionTable in our repo?)  But moreso I would like to see most of this code removed in favor of just an interface that gets a full URL with all the params, and we then send that off to the server with a .csv extension..

src/plugins/data_table/data_table.js
```diff
@@ -190,6 +200,7 @@ var LocalDataTable = (function() {
         paginateLengthMenu : [ 10, 25, 50, 100 ],
         paginateLength : 10,
         selectedIds : [],
+        filteringEnabled: true,
```
nburwell: do we want filteringEnabled as a default?

dist/backdraft.js
colin-allen: didn't know you could do that with jQuery!  awesome.  :smile:

dist/backdraft.js
colin-allen: sure

dist/backdraft.js
colin-allen: the issue is the event.target may be a child object of the filterMenu.  I can attach a class to all of the items in the filterMenu, but it will look extremely redundant and ugly in the DOM.  there's probably a better way to do this.

dist/backdraft.js
colin-allen: Yes on naming it activeFilterMenu.  filterActive is a class used to indicate what image is used for the filter icon (when filters are set for a column vs. when there are no filters set for it).  I'll see if I can clarify these class names.  activeFilterMenu seems good, not sure what to use instead of filterActive/filterInactive.

dist/backdraft.js
colin-allen: This is just a boolean that I use for logic later on.  will rename as per your recommendation.

src/plugins/data_table/data_table.js
colin-allen: Maybe?  I'm not sure.  I'm happy to set this to false instead.

dist/backdraft.js
colin-allen: Nesting sounds awesome.  Will do.  I think the Hungarian notation there is just because I didn't want a variable conflict and wanted to clarify the difference between the base string used to make the template and the template itself.  (e.g. ``sFilterStringTemplate`` and ``filterStringTemplate``)  Is there a more preferred naming to do to tell the difference between these, or would you be happy with just the nesting?

Will use className.  Much better.  :)

dist/backdraft.js
colin-allen: that sounds great, thanks.

dist/backdraft.js
colin-allen: yes, will use temporary variables and comments to clarify the code and look into extracting the logic.

dist/backdraft.js
colin-allen: well "Clear" isn't a cancel button, it removes the active filters.  

* Both "Clear" and "Filter" should close the menu
* Opening a new menu should close the menu
* Clicking on the toggle button should toggle the menu
* Clicking out of the menu should close it.
** Else the user has to click filter, clear, or the toggle button.  This doesn't cover the use case where the user wants to apply filters to multiple columns before submitting the filter request to the server and isn't aware they can click the toggle button to make the menu disappear, which isn't always the case.

dist/backdraft.js
colin-allen: ok, I'll see if I can make a url that does what we want.

dist/backdraft.js
colin-allen: ok, will try doing that.

dist/backdraft.js
nburwell: if you only care if the click is on the filtermenu or child elements, there are definitely ways to do that.  I forget the exact syntax but google or jquery event click docs should have examples.. basically you want to check if your event.target is in the DOM hierarchy of X element (where X I think is the filtermenu, found from a specifically named class I hope)

nburwell:
@colin-allen last question on this PR:

Where are the specs (JS unit tests) for the filter changes? These need to be added at the backdraft level (then some at the reporting app level as more of an integration test to show it fully plumbed and at least sending a new ajax request off to server)

colin-allen:
@nburwell I still want to try to address the two pending suggestions before going through with this PR.  (using backdraft views and not using ajax for exporting to CSV).

As far as tests are concerned, I wasn't entirely certain what sort of additional test coverage I wanted to make with these changes, though I am getting a better idea now.

I think there should be some tests associated with 
* How filter menus are shown/hidden (the conditions)
* How the toggle filter icon changes
* What happens when a user clicks the Filter/Clear buttons

colin-allen:
Now using a view for filters, much nicer :smile: 

Pending items:
* Tests
* _fetchCSV without ajax magic

colin-allen:
Yay tests!

colin-allen:
Ok so, I implemented _fetchCSV without doing an ajax post and stuff (what I had before), but this has the exact limitation which is the reason why I did the ajax stuff.  It doesn't pay attention to user column order, columns hidden by the user, or filters applied by the user.  If that's the functionality you want then setting window.location to the .csv address is all we need, I just didn't think that was our intended functionality for exporting.  (I looked at EXT reports and this is also the case there)

src/plugins/data_table/data_table.js
```diff
@@ -2,30 +2,167 @@ var LocalDataTable = (function() {
...
+  };
+
+  var DataTableFilter = Base.View.extend({
+    filterTemplate : _.template('\
+      <div class="toggle-filter-button" data-toggle="dropdown">\
+        <span class="<%= filterButtonClass %>"></span>\
+      </div>\
+    ', null, DEFAULT_JST_DELIMS),
+
+    filterMenuTemplate : _.template('\
```
InvokeJLa: I would break up these filters like you had before. So instead of a single template that handles all filters, there would be a string one, list, numeric, etc.

src/plugins/data_table/data_table.js
```diff
@@ -2,30 +2,167 @@ var LocalDataTable = (function() {
...
+      <div class="filterMenu dropdown-menu <%= listClass %>">\
+        <% if (filter.type === "string") { %>\
+          <input class="filter-string" id ="value" type="text" placeholder="Search <%= title %>" />\
```
InvokeJLa: an id should be unique to the entire page. If so if there is more than 1 element active, there would be a conflict. Can we give it a more descriptive id? Like, filter-string-[name of column] or something?

src/plugins/data_table/data_table.js
```diff
@@ -2,30 +2,167 @@ var LocalDataTable = (function() {
...
+        <% } else if (filter.type === "numeric") { %>\
+          <ul>\
+            <li> &gt; <input id="gt" class="filter-numeric" type="text" /></li> \
+            <li> &lt; <input id="lt" class="filter-numeric" type="text"/></li> \
+            <li> = <input id="eq" class="filter-numeric" type="text" /></li> \
```
InvokeJLa: Same id issue with these and the one below it.

src/plugins/data_table/data_table.js
```diff
@@ -2,30 +2,167 @@ var LocalDataTable = (function() {
...
+          </ul>\
+        <% } else if (filter.type === "list") { %>\
+          <ul>\
+            <% _.each(filter.options, function(element, index) { %>\
+              <li>\
+                <label>\
```
InvokeJLa: For better usability, and in case you need to programmatically check a checkbox later, it's generally best practice to wrap checkboxes and their individual labels (like you did here). But also to give it a markup like this:

<label for="someUniqueId"><input type="checkbox" id="someUniqueId" name="somename" />Text that is displayed to the user.</label>

Without the "label for", the browser is interpreting that "this label is for the input box between the label tags." Adding the "for" attribute explicitly tells the browser, "this label is for this checkbox."

src/plugins/data_table/data_table.js
```diff
@@ -2,30 +2,167 @@ var LocalDataTable = (function() {
...
+      var listClass = "";
+      if (this.filter.type === "list") {
+        if (this.filter.options.length > 30) {
+          listClass = " triple";
+        } else if (this.filter.options.length > 15) {
+          listClass = " double";
+        } else {
+          listClass = " single";
+        }
+      }
```
InvokeJLa: I _think_ if we move the display logic in the wrapper template like I suggested above, you can remove this portion.

src/plugins/data_table/data_table.js
```diff
@@ -2,30 +2,167 @@ var LocalDataTable = (function() {
...
+          filter: this.filter,
+          title: this.title,
+          attr: this.attr,
+          listClass: listClass
+        })
+      );
+    },
+
+    _onToggleClick: function(event) {
+      var table = this.table;
```
InvokeJLa: Is there a reason to alias this.table? There doesn't look like there are any scope conflicts in the rest of this function.

src/plugins/data_table/server_side_data_table.js
```diff
@@ -88,6 +88,71 @@ var ServerSideDataTable = (function() {
...
+                a.click();
+              }
+            } else {
+              window.location = downloadUrl;
+            }
+
+            setTimeout(function () { URL.revokeObjectURL(downloadUrl); }, 100); // cleanup
+          }
+        }
+      })*/
```
InvokeJLa: I do this a lot too -- keep stuff around just in case I need it in the future. But since we have source control, would you please remove this?

src/plugins/data_table/data_table.js
colin-allen: I just needed somewhere to store that string value in the DOM.  It doesn't have to be the ID, I can put it in a data-* attributes.

src/plugins/data_table/server_side_data_table.js
colin-allen: Well the main reason I commented this here is because I want to know if the functionality without it is what we intended, as per my comment yesterday before b109198.  I will definitely delete it if we're good on that.

src/plugins/data_table/data_table.js
colin-allen: so label for requires me to link to the ID of the checkbox element, so I'll have to make a unique id for each checkbox element.  ok.

src/plugins/data_table/data_table.js
colin-allen: I kind of liked having the filters all in a single template and having the menu separate.  What is the reasoning for breaking out individual filters instead of the entire filter menu?

spec/plugins/data_table/server_side.js
```diff
@@ -480,54 +481,184 @@ describe("DataTable Plugin", function() {
...
+              $(".filterMenu #eq", this).val("3").trigger("change");
+              $(".btn-clear", this).click();
+              expect($(".filterMenu #eq", this).val()).toEqual("");
+            }
+            else if (col.filter.type === "list") {
+              $(".filterMenu #value", this).prop("checked", true).trigger("change");
+              $(".btn-clear", this).click();
+              $(".filterMenu #value", this).prop("checked", false).trigger("change");
+              expect($(".filterMenu #value", this).prop("checked")).toEqual(false);
+            }
```
InvokeJLa: In general, on all of these tests tend to check if the table wrapper exists, then check if it has a column, and a filter, etc etc. Instead, we should setup the table/filter assuming that those are there. Then allow the "expect" assertions to go through and run.

I understand that there are cases where columns would not have filters. If that's the case, I'd create different test data where all columns have filters. That would eliminate a lot of the checking.

src/plugins/data_table/data_table.js
InvokeJLa: Makes it easier to read and in the future debug/modify if they're in separate places.

src/plugins/data_table/data_table.js
```diff
@@ -317,142 +451,55 @@ var LocalDataTable = (function() {
...
-              filter[this.id].splice(index, 1);
-            if (filter[this.id].length === 0)
-              filter[this.id] = null;
-          }
-        } else if (this.value === "") {
-          filter[this.id] = null;
-        } else {
-          filter[this.id] = this.value;
+      // Close active filter menu if user clicks on document
+      $(document).click(function (event) {
```
InvokeJLa: Gah. Not my favorite way to handle this, but I don't know of a better solution.

InvokeJLa:
CR done for now. Talked to @colin-allen about the general change to the tests.

nburwell:
> Ok so, I implemented _fetchCSV without doing an ajax post and stuff (what I had before), but this has the exact limitation which is the reason why I did the ajax stuff. It doesn't pay attention to user column order, columns hidden by the user, or filters applied by the user. If that's the functionality you want then setting window.location to the .csv address is all we need, I just didn't think that was our intended functionality for exporting. (I looked at EXT reports and this is also the case there)

That's not the functionality I want. It should include all 3 of those.  Isn't the server using the report settings object when exporting to CSV? If not, it should and we'd get the order and visibility of columns automatically, right? Then on the filter piece, that should already be supported in EXT, and is critical to plumb through for export. BTW we already need the filters to be updating the URL and allowing you to restore a report via browser-refresh to the same filters you had.  Does that work yet? If not, please add a to do item to track that.  And then, once that is done, the same plumbing / helper method to generate that URL structure for filters should be used when clicking export was my thought.

I'm back in office so we can touch base today if need be on these pieces. Thanks.

src/plugins/data_table/data_table.js
nburwell: I would rather not have an ID for each checkbox element if just for the label here..  besides the extra code and work, there's a chance of not actually having unique IDs for the entire DOM on page (which would be a worse HTML bug IMO).  And how you wrapped it here works for me (since there is only one input within the `label` I don't see any ambiguity and the browser is going to have the same behavior here).

src/plugins/data_table/data_table.js
nburwell: did you mean `<%= value %>` up above, because right now weren't you just setting id to the literal string `value`?  Trying to figure out exactly what you needed to persist. I agree a data-* attribute would be much better. I would also name the attribute something clear to indicate what it is storing.

src/plugins/data_table/data_table.js
colin-allen: ok, sounds good to me!  I like less work.  :stuck_out_tongue_winking_eye:

src/plugins/data_table/data_table.js
colin-allen: This can be integrated into the template even with the current template, it's just ugly.

spec/plugins/data_table/server_side.js
colin-allen: I think making different test data would be best, then I can test each column manually.

src/plugins/data_table/data_table.js
colin-allen: I understand how document event handlers are messy, but there's really no better way that I'm aware of with dropdown-menus, at least.

colin-allen:
I'm going to vote we leave exporting as an AJAX post for now, because making it use url params and report settings is a considerably large chunk of work that we really can't afford to do if we want this story to be completed any time soon.

The functionality is equivalent, so I don't see any reason we shouldn't just roll with this for the time being.

nburwell:
okay sounds good. Please track somewhere that filters are not in the URL
(therefore not bookmarkable or shareable), and that is a feature we lost
from Ext.  And could also note on that item that we could reconsider export
at that time.

Thanks!

colin-allen:
@nburwell will do!  Is this :+1: ?

nburwell:
Where's my list of review questions and whether they have been resolved or not ;)

Taking a look again..

colin-allen:
The only thing I'm aware of that isn't here is moving it out into a separate file.  I tried that but it made literally every test fail, so I think I did something wrong.  Not sure what though.  :neutral_face:

src/plugins/data_table/data_table.js
```diff
@@ -2,30 +2,249 @@ var LocalDataTable = (function() {
...
+    },
+
+    _onClearClick: function() {
+      $("input[type=text]", this.head).val("");
+      $("input[type=checkbox]", this.head).attr("checked", false);
+      $("input", this.head).trigger("change");
+      this.table.dataTable._fnAjaxUpdate();
+    }
+  });
+
```
nburwell: is it easy to put all the above in a separate .js file like we discussed? I hope we can follow the patterns set by other files like column_manager.js

nburwell:
One last hope and desire, and if you can move it to a separate file, then :+1:

nburwell:
Did you add a line like this? https://github.com/Invoca/backdraft/blob/master/src/plugins/data_table/index.js#L10-10

colin-allen:
@nburwell yes I did, and that line is what broke it.  I'm not sure why, will dork around with it a bit and slack you if I can't get it working in an hour or something.

colin-allen:
Alright, we're good!

nburwell:
:+1:

PR #30, opened 2015-08-09
Add simple api param support

nburwell:
* Bumps underscore version to 1.8.3
* Provides a new option when declaring a DataTable view

    ```
    app.view.dataTable("T", {
      rowClassName : "R",
      serverSide : true,

      /* new option: */
      simpleParams : true
    });
    ```

When provided, rather than an ajax url like this when the table loads / refreshes:
```
/somewhere?sEcho=2&iColumns=4&sColumns=&iDisplayStart=0&iDisplayLength=10&mDataProp_0=function&mDataProp_1=function&mDataProp_2=function&mDataProp_3=function&sSearch=&bRegex=false&sSearch_0=&bRegex_0=false&bSearchable_0=false&sSearch_1=&bRegex_1=false&bSearchable_1=true&sSearch_2=&bRegex_2=false&bSearchable_2=true&sSearch_3=&bRegex_3=false&bSearchable_3=false&iSortCol_0=0&sSortDir_0=desc&iSortingCols=1&bSortable_0=true&bSortable_1=true&bSortable_2=true&bSortable_3=false&_=1439129476800
```

We get this:
```
/somewhere?sort_by=name&sort_dir=desc&limit=10&start=0&request_id=2&monkey=chicken&_=1439131967030
```

The big thing being that sort_by is the sort param based on the column attribute, and otherwise the other paging / sorting params are simply renamed, and the other column metadata params removed


review-ninja:
<a href="http://app.review.ninja/Invoca/backdraft/pull/30" target="_blank">ReviewNinja

nburwell:
@InvokeJLa can you please review?

src/plugins/data_table/server_side_data_table.js
```diff
@@ -70,13 +70,49 @@ var ServerSideDataTable = (function() {
 
     // dataTables callback to allow addition of params to the ajax request
     _addServerParams : function(aoData) {
+      if (this.simpleParams) {
+        var sortBy, sortDir, limit, start, requestId;
+
+        var indexOfSortedColumn = this._getDataTableParam(aoData, "iSortCol_0");
```
InvokeJLa: Wouldn't this limit the number of columns it can sort by default to 1? Even though our current Ext lib is limited to 1, DataTables can do multi-sort. Do we want to hard code this to 1 for Simple Params.

nburwell: good question, yes we do lose that flexibility. I'll look into a nice way to extend this to support arbitrary sort columns

src/plugins/data_table/server_side_data_table.js
```diff
@@ -70,13 +70,49 @@ var ServerSideDataTable = (function() {
...
+        requestId = this._getDataTableParam(aoData, "sEcho");
+
+        // clear out existing array (but keeping reference to existing object)
+        aoData.splice(0, aoData.length);
+
+        aoData.push({ name: "sort_by",    value: sortBy });
+        aoData.push({ name: "sort_dir",   value: sortDir });
+        aoData.push({ name: "limit",      value: limit });
+        aoData.push({ name: "start",      value: start });
+        aoData.push({ name: "request_id", value: requestId });
```
InvokeJLa: Is request_id used somewhere else? Or is it meant as like a memo field on a check?

src/plugins/data_table/server_side_data_table.js
```diff
@@ -70,13 +70,49 @@ var ServerSideDataTable = (function() {
...
       for (var key in this._serverParams) {
         aoData.push({ name : key, value : this._serverParams[key] });
       }
-      // add column attribute mappings as a parameter
-      _.each(this._columnManager.columnAttrs(), function(attr) {
-        aoData.push({ name: "column_attrs[]", value: attr });
-      });
+    },
+
+    _getDataTableParam : function(data, key) {
```
InvokeJLa: The method gets a value if it exists, or returns null. Nit picking on my part, and I didn't see anything anything in our style guide about naming something like this. And do we want to allow something like this since on a server-side table, some of these are required.

spec/plugins/data_table/server_side.js
```diff
@@ -290,15 +299,43 @@ describe("DataTable Plugin", function() {
...
+      expect(jasmine.Ajax.requests.mostRecent().url).toMatch("iSortCol_0=1");
+      expect(jasmine.Ajax.requests.mostRecent().url).toMatch("sSortDir_0=desc");
+      expect(jasmine.Ajax.requests.mostRecent().url).toMatch("sEcho=2");
+
+      expect(jasmine.Ajax.requests.mostRecent().url).not.toMatch("sort_by");
+      expect(jasmine.Ajax.requests.mostRecent().url).not.toMatch("sort_dir");
+      expect(jasmine.Ajax.requests.mostRecent().url).not.toMatch("request_id");
+    });
+
+    it("should pass simpler paging and sort param names when in simple_params mode", function() {
```
InvokeJLa: Since sorting is optional, and it's covered by a different block from the standard full-length report, would want want to test a case where sorting isn't sent? And that both sort_by and sort_dir are present, not one or the other? Or maybe default a sort dir if not specified?

InvokeJLa:
@nburwell CR done. :smile:  I had some small questions about it, otherwise, it looks good.

src/plugins/data_table/server_side_data_table.js
nburwell: as discussed we are leaving this for now.. if you need multi-sort, don't set simpleParams (and if & when that is coming up and we want simpleParams + multisort we can take on then)

src/plugins/data_table/server_side_data_table.js
nburwell: as part of further cleanup and some bug fixes, this is not `_getDataTableParamIfExists`

nburwell: I think all of them are optional which is nice. Server can choose reasonable defaults for each

spec/plugins/data_table/server_side.js
nburwell: Added test cases for no sort at all, and no sort name but passing a sort dir.
Also fixed a bug around sorting on first column (index 0, so the `if (0)` didn't work so well..)

nburwell:
Ready for re-review. BTW I had forgotten to add the response param mapping too! So, that's in with some crazy tests that I did TDD on so despite the crazy stubs, they do work =)

spec/plugins/data_table/server_side.js
```diff
@@ -290,15 +306,238 @@ describe("DataTable Plugin", function() {
...
+      table = new app.Views.AjaxMethodTestTable({collection: collection});
       table.render();
 
       expect(jasmine.Ajax.requests.mostRecent().method).toEqual("GET");
     });
+
+    it("should handle DT 1.10 param for sEcho (draw) for ajax response", function() {
+      table = new app.Views.T({ collection : collection });
+
+      var callbackSpy = jasmine.createSpy("callbackSpy");
```
InvokeJLa: This block seems to be repeated throughout the some of the tests. If possible, can we move it out? However, it's only 4 lines, and dealing with the "this" context can get convoluted, so I can live with it repeating in the tests.

nburwell: Ended up not being that bad. Shared the logic: debf41c

InvokeJLa:
Yay :+1:

PR #31, opened 2015-08-09
Test Removal of Review Ninja - Delete when desired.

sherriman:


nburwell:
looks good (that review ninja is not enbled)

Issue #1, opened 2014-03-16
Add jQuery proxy helpers to views


euge:
- html
- hide
- show
- maybe some others like append

PR #33, opened 2015-08-26
87/t/STORY-2792_backdraft_detail_reports: Fixes bug in dataTables

Vicentecarrillo:
outerText method doesn't work in firefox, bug was causing dataTables to skip the rendering of filters. Replaced with textContent method which is functionally equivalent and works in firefox.

[Review on Reviewable](https://reviewable.io/reviews/invoca/backdraft/33)

### DIFF
--- a/src/plugins/data_table/data_table.js
+++ b/src/plugins/data_table/data_table.js
@@ -377,7 +377,8 @@ var LocalDataTable = (function() {
       table.dataTable.find("thead th").each(function (index) {
         // here we use the text in the header to get the column config by title
         // there isn't a better way to do this currently
-        var title = this.textContent;
+        debugger;
+        var title = this.textContent || this.innerText;
         var col = cg.columnConfigByTitle.attributes[title];
 
         if (col) {
--- a/src/plugins/data_table/data_table.js
+++ b/src/plugins/data_table/data_table.js
@@ -377,7 +377,7 @@ var LocalDataTable = (function() {
       table.dataTable.find("thead th").each(function (index) {
         // here we use the text in the header to get the column config by title
         // there isn't a better way to do this currently
-        var title = this.outerText;
+        var title = this.textContent;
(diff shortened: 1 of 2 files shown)


Vicentecarrillo:
@nburwell @melch can I get some CR?

Vicentecarrillo:
@InvokeJLa CR?

InvokeJLa:
@Vicentecarrillo Starting CR

src/plugins/data_table/data_table.js
```diff
@@ -377,7 +377,7 @@ var LocalDataTable = (function() {
       table.dataTable.find("thead th").each(function (index) {
         // here we use the text in the header to get the column config by title
         // there isn't a better way to do this currently
-        var title = this.outerText;
+        var title = this.textContent;
```
InvokeJLa: I'm concerned about this change.

1) The original code assigns outerText to ``title``. ``outerText`` includes the contents of the tag, AND the surrounding tags themselves. 

An example can be found here: http://stackoverflow.com/questions/18481382/what-is-the-difference-between-innertext-and-outertext

https://msdn.microsoft.com/en-us/library/ms534311(v=vs.85).aspx

So with this change, whatever tags that surround the text will not get pulled into ``title``.

2) ```outerText``` is an IE specific tag, and it was only copied in the impl. of Opera, Chrome, and I *think* Safari. So the original code shouldn't have used it exclusively. However, ```textContent``` is not supported by IE (Maybe in MS Edge, but I'm not sure).

If we're sure that titles never have any HTML in them (see #1 above), the cross browser way to do this is: ```x.textContent || x.innerText```

InvokeJLa:
I think there's an accompanying test with this?

InvokeJLa:
@Vicentecarrillo CR done.

InvokeJLa:
:+1: For me.

PR #36, opened 2015-09-04
STORY 2678 modify fetch_csv: removes aoData from params to fetchCSV, …

trevorfrese:
…adds error handling for goToWindowLocation

@InvokeJLa can I get a CR?


InvokeJLa:
Wait this is just a updating the auto generated distribution file, right? Because the previous PR had this in the source files?

trevorfrese:
Crap. You're right I didn't realize that it was in the last PR. Sorry about that. Yes this would only update the dist. Is it worth it to keep dist up to date?

InvokeJLa:
Yeah, we need to keep the dist file updated since that's what goes out to apps that use backdraft.

InvokeJLa:
:+1:

PR #37, opened 2015-10-08
UX :: column resize support

nburwell:
@mileszim @albertovillalobos @Vicentecarrillo 

I looked into resize plugins for dataTables and this is the best (only?) one I could find that worked with our older version of dataTables and also is just a fork of the colReorder plugin we already use, so the changes are fairly minimal.

I first checked in the new ColReorderWithResize plugin as is: https://github.com/Invoca/backdraft/commit/4ac17c5a966f76c3b6659591299672a782fa5a04

Then I applied my fixes/tweaks on top of it that are needed to get this working for full-width grids.

Disclaimer: it is not working out of the box yet on tables that are fixed within a viewport. Some tweaks will need to be made before we can enable safely on the management grids. (That is why it's disabled by default).

I'd like to get this in and start playing with it on the reporting app (but we need to be sure this does not break the basic ColReorder functionality that is already in use in many places).

The diff isn't shown in github on `src/plugins/data_table/dataTables.colReorder.js` lame. You may want to try using reviewable.io or diffing locally.

### DIFF
--- a/spec/plugins/data_table/general.js
+++ b/spec/plugins/data_table/general.js
@@ -557,13 +557,57 @@ describe("DataTable Plugin", function() {
       expect(reorderableSpy).not.toHaveBeenCalled();
     });
 
+    it("should default column resize to off", function() {
+      app.view.dataTable.row("abc", {
+        columns : [
+          { attr: "name", title: "Name" },
+          { attr: "age", title: "Age" }
+        ]
+      });
+
+      app.view.dataTable("def", {
+        rowClassName : "abc"
+      });
+
+      var table = new app.Views.def({ collection : collection }).render();
+
+      expect(table.resizableColumns).toEqual(false, "Table resizableColumns default");
+      expect(table._colReorder.s.allowResize).toEqual(false, "ColReorder allowResize");
+      expect(table._colReorder.s.allowReorder).toEqual(true, "ColReorder allowReorder");
+    });
+
+    it("should allow columns to be resized with flag", function() {
+      app.view.dataTable.row("abc", {
+        columns : [
+          { attr: "name", title: "Name" },
+          { attr: "age", title: "Age" }
+        ]
+      });
+
+      app.view.dataTable("def", {
+        rowClassName : "abc",
+        resizableColumns : true
+      });
+
+      var table = new app.Views.def({ collection : collection }).render();
+
+      expect(table.resizableColumns).toEqual(true, "Table resizableColumns default");
+      expect(table._colReorder.s.allowResize).toEqual(true, "ColReorder allowResize");
+      expect(table._colReorder.s.allowReorder).toEqual(true, "ColReorder allowReorder");
+
+      // these should set to off as they cause display issues
+      expect(table._colReorder.s.bAddFixed).toEqual(false, "ColReorder bAddFixed");
+      expect(table._colReorder.s.bResizeTableWrapper).toEqual(false, "ColReorder bResizeTableWrapper");
+      expect(table._colReorder.s.allowHeaderDoubleClick).toEqual(false, "ColReorder allowHeaderDoubleClick");
+    });
+
     it("should keep track of columns being reordered", function() {
       app.view.dataTable.row("abc", {
         columns : [
           { attr: "name", title: "Name" },
           { attr: "age", title: "Age" },
           { attr: "location", title: "Location" },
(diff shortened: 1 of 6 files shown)


Vicentecarrillo:
@nburwell I think this looks promising, especially given that it's a fork of existing code we already use and the fact that we can turn it on/off. All for getting this in and seeing if we can get it working 100% :+1:

Vicentecarrillo:
Reviewed 6 of 6 files at r1.
Review status: all files reviewed at latest revision, 8 unresolved discussions, all commit checks successful.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/invoca/backdraft/37)

albertovillalobos:
@nburwell I've merged latest master in

nburwell:
Looks good to me

spec/plugins/data_table/general.js
```diff
@@ -557,13 +557,57 @@ describe("DataTable Plugin", function() {
...
+
+      var table = new app.Views.def({ collection : collection }).render();
+
+      expect(table.resizableColumns).toEqual(true, "Table resizableColumns default");
+      expect(table._colReorder.s.allowResize).toEqual(true, "ColReorder allowResize");
+      expect(table._colReorder.s.allowReorder).toEqual(true, "ColReorder allowReorder");
+
+      // these should set to off as they cause display issues
+      expect(table._colReorder.s.bAddFixed).toEqual(false, "ColReorder bAddFixed");
+      expect(table._colReorder.s.bResizeTableWrapper).toEqual(false, "ColReorder bResizeTableWrapper");
```
HParker: Looks like we might want to increase the max line length?

nburwell: those are from long ago (Oct 8), and we discussed it back then. I thought you had turned off line length altogether based on that.  I haven't seen a hound error about line length since (but we haven't done much dev here since either)

HParker: ah, thanks! I saw the recent comment, but didn't realize it is an old hound run.

albertovillalobos:
@nburwell Browser compatibility fixes done, matches the one on web repo (i double checked).

src/plugins/data_table/data_table.js
```diff
@@ -271,6 +276,10 @@ var LocalDataTable = (function() {
       this._columnManager.on("change:visibility", this._onColumnVisibilityChange);
       this._columnManager.applyVisibilityPreferences();
       if (this.collection.length) this._onReset(this.collection);
+      // if resizeable, add resizeable class
+      if (this._colReorder.s.allowResize) {
+        this.$('table').addClass('resizeableColumns');
```
nburwell: can you namespace this class please? `dataTable-resizeableColumns` ?

nburwell: Also can you add a test for this

src/plugins/data_table/data_table.js
```diff
@@ -271,6 +276,10 @@ var LocalDataTable = (function() {
       this._columnManager.on("change:visibility", this._onColumnVisibilityChange);
       this._columnManager.applyVisibilityPreferences();
       if (this.collection.length) this._onReset(this.collection);
+      // if resizeable, add resizeable class
+      if (this._colReorder && this._colReorder.s.allowResize) {
+        this.$("table").addClass('dataTable-resizeableColumns')
```
nburwell: weird that hound didn't comment on this line (or maybe it did and it's been deleted). But please keep the quoting consistent for now with double quotes

spec/plugins/data_table/general.js
```diff
@@ -559,13 +559,91 @@ describe("DataTable Plugin", function() {
...
+      expect(table._colReorder.s.allowHeaderDoubleClick).toEqual(false, "ColReorder allowHeaderDoubleClick");
+    });
+
+    it('should add dataTable-resizeableColumns class to resizable tables', function() {
```
nburwell: I think it's worth addressing these hound comments about double & single mixed.  We have used double throughout this code base

src/plugins/data_table/data_table.js
HParker: funny it seems to catch inconsistent quotes in a file, but not on a line.

nburwell:
some quote comments but otherwise :+1:

PR #38, opened 2015-10-08
add hound config

HParker:
@nburwell here is a first stab at doing some hound config. I think this might be a bit of trial and error. Hopefully it helps.


nburwell:
:+1:

PR #39, opened 2015-10-30
WEB-1492 Checkbox "column" is moveable in backdraft Manage pages

codingiam:


src/plugins/data_table/column_config_generator.js
```diff
@@ -31,6 +31,11 @@ var ColumnConfigGenerator =  Backdraft.Utils.Class.extend({
       }
     });
 
+    // make the bulk column the first one if present
+    this.columnsConfig = _.sortBy(this.columnsConfig, function (columnConfig) {
```
nburwell: looks like we need to tweak HoundCI to know about some global variables. @HParker can you point me to where we can do that?

dist/backdraft.js
```diff
@@ -527,6 +527,11 @@ _.extend(Plugin.factory, {
       }
     });
 
+    // make the bulk column the first one if present
+    this.columnsConfig = _.sortBy(this.columnsConfig, function (columnConfig) {
+      return !columnConfig.bulk;
+    });
```
codingiam: @nburwell With this the bulk column gets moved in the first position.

dist/backdraft.js
```diff
@@ -1252,7 +1257,10 @@ _.extend(Plugin.factory, {
           self._columnManager.columnsSwapped(fromIndex, toIndex);
           // pass event up
           self._onColumnReorder();
-        }
+        },
+        // iFixedColumns configures how many columns should be unmovable starting from left
+        // if the first column is the bulk column we make it unmovable
+        iFixedColumns: this.$el.find("th:first.bulk :checkbox").length
```
codingiam: @nburwell We use this selector in other places so I think it should be safe to use for this situation.

dist/backdraft.js
nburwell: ah, cool good to know

dist/backdraft.js
nburwell: fair enough.. what about making the `th.first.bulk` a "constant", ie `this.BULK_COLUMN_SELECTOR`  and using throughout?

codingiam: I'll do that.

src/plugins/data_table/column_config_generator.js
HParker: https://houndci.com/configuration
https://github.com/Invoca/backdraft/blob/master/.hound.yml

nburwell: Thanks, added to master. Going to merge in here and see if we get any hounding

codingiam:
This is ready for rereview.

nburwell:
:+1:

PR #43, opened 2015-11-10
91/t/story 2926 column filtering

mileszim:
https://ringrevenue.atlassian.net/browse/STORY-2926

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -36,7 +36,9 @@ module.exports = function(grunt) {
             "vendor/underscore-1.8.3.js",
             "vendor/backbone-1.1.2.js",
             "vendor/jquery.dataTables-1.9.4.js",
-            "vendor/mock-ajax.js"
+            "vendor/mock-ajax.js",
+            "vendor/bootstrap-3.1.1-dist/js/bootstrap.js",
+            "vendor/popover_menu.js"
           ],
           specs : [
             "spec/**/*.js"
--- a/src/plugins/data_table/filter_view.js
+++ b/src/plugins/data_table/filter_view.js
@@ -46,8 +46,18 @@ var DataTableFilter = (function(options) {
 
       this.afterRender();
 
+      // Bind button clicks
       this.$('.btn-filter').click(this.parentView._onFilterClick.bind(this.parentView));
       this.$('.btn-clear').click(this.parentView._onClearClick.bind(this.parentView));
+
+      // Bind "enter" key
+      this.$('.filterMenu').keyup(function(event) {
+        var key = event.keyCode || event.which;
+        if (key === 13) {
+          this.parentView._onFilterClick.call(this.parentView);
+        }
+      }.bind(this));
+      
       return this;
     },
 
@@ -269,6 +279,8 @@ var DataTableFilter = (function(options) {
     },
 
     _onToggleClick: function (event) {
+      // close popoverMenus except for this one
+      $(".toggle-filter-button").not(event.currentTarget).popoverMenu('hide');
       event.stopImmediatePropagation();
     },
 
--- a/src/plugins/data_table/filter_view.js
+++ b/src/plugins/data_table/filter_view.js
@@ -80,24 +80,42 @@ var DataTableFilter = (function(options) {
 
   var NumericFilterMenu = DataTableFilterMenu.extend({
     filterMenuClass: "filterMenu-numeric",
-    tagName: "ul",
-
     menuTemplate: _.template('\
-        <li> &gt; <input data-numeric-filter-name="gt" class="filter-numeric filter-numeric-greater" type="text" /></li> \
-        <li> &lt; <input data-numeric-filter-name="lt" class="filter-numeric filter-numeric-less" type="text"/></li> \
-        <li> = <input data-numeric-filter-name="eq" class="filter-numeric filter-numeric-equal" type="text" /></li> \
-      ', null, DEFAULT_JST_DELIMS),
+      <div class="filter-text">Show items with value that:</div>\
+      <select class="filter-type" data-filter-id="first-filter">\
+        <option selected value="gt">is greater than</option> \
+        <option value="lt">is less than</option> \
+        <option value="eq">is equal to</option> \
+      </select> \
+      <input id="first-filter" class="filter-value" type="text" data-filter-type="gt"  /> \
+      <div class="filter-text">and</div> \
+      <select class="filter-type" data-filter-id="second-filter">\
+        <option value="gt">is greater than</option> \
+        <option selected value="lt">is less than</option> \
+        <option value="eq">is equal to</option> \
+      </select> \
+      <input id="second-filter" class="filter-value" type="text" data-filter-type="lt" /> \
+    ', null, DEFAULT_JST_DELIMS),
 
     _onInputChange: function (event) {
-      var filterInput = event.target;
-      var numericValueName = $(filterInput).attr("data-numeric-filter-name");
-      if (filterInput.value === "") {
-        this.filter[numericValueName] = null;
+      event.stopImmediatePropagation();
+      var filterType = event.target.getAttribute('data-filter-type'),
+          filterValue = event.target.value;
+      if (filterValue === "") {
+        this.filter[filterType] = null;
         this.parentView._toggleIcon(false);
       } else {
-        this.filter[numericValueName] = filterInput.value;
+        this.filter[filterType] = filterValue;
         this.parentView._toggleIcon(true);
       }
+    },
+
+    afterRender: function() {
+      this.$('.filter-type').select2().bind('change', function(event) {
+        var filterElementId = event.target.getAttribute('data-filter-id'),
+          filterType = event.target.value;
+        $('#'+filterElementId).attr('data-filter-type', filterType);
+      });
     }
   });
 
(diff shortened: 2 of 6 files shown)


mileszim:
@nburwell this should be ready for CR.

Gruntfile.js
```diff
@@ -36,7 +36,9 @@ module.exports = function(grunt) {
             "vendor/underscore-1.8.3.js",
             "vendor/backbone-1.1.2.js",
             "vendor/jquery.dataTables-1.9.4.js",
-            "vendor/mock-ajax.js"
+            "vendor/mock-ajax.js",
+            "vendor/bootstrap-3.1.1-dist/js/bootstrap.js",
```
nburwell: really wanting this (filtering) to be a separate, optional plugin in a backdraft-plugins repo. hate pulling BS in as a dependency.

vendor/popover_menu.js
```diff
@@ -0,0 +1,74 @@
++function ($) {
```
nburwell: should we stick this in a shared repo (ux-components one?) and have it be our first thing that we pull into both web and backdraft?

Otherwise how do we prevent these from drifting

src/plugins/data_table/filter_view.js
```diff
@@ -56,7 +56,15 @@ var DataTableFilter = (function(options) {
 
   var StringFilterMenu = DataTableFilterMenu.extend({
     menuTemplate: _.template('\
-        <input class="filter-string" id ="value" type="text" placeholder="Search <%= title %>" />\
+      <div class="filterMenu filterMenu-string">\
+        <div class="filter-text">Show:</div>\
+        <div class="icon-addon addon-sm">\
+          <input class="filter-string form-control input-sm" id="value" type="text" name="filter-string" />\
```
nburwell: can you remove `id="value"` please? i know existed before but there will be lots of these on the page, and w/o a unique ID

src/plugins/data_table/filter_view.js
```diff
@@ -68,43 +76,79 @@ var DataTableFilter = (function(options) {
...
   var NumericFilterMenu = DataTableFilterMenu.extend({
-    tagName: "ul",
-
     menuTemplate: _.template('\
-        <li> &gt; <input data-numeric-filter-name="gt" class="filter-numeric filter-numeric-greater" type="text" /></li> \
-        <li> &lt; <input data-numeric-filter-name="lt" class="filter-numeric filter-numeric-less" type="text"/></li> \
-        <li> = <input data-numeric-filter-name="eq" class="filter-numeric filter-numeric-equal" type="text" /></li> \
-      ', null, DEFAULT_JST_DELIMS),
+      <div class="filterMenu filterMenu-numeric">\
+        <div class="filter-text">Show items with value that:</div>\
```
nburwell: Grammar seems off here.. maybe "Show items with a value that" ?

src/plugins/data_table/filter_view.js
```diff
@@ -56,7 +56,15 @@ var DataTableFilter = (function(options) {
...
     menuTemplate: _.template('\
-        <input class="filter-string" id ="value" type="text" placeholder="Search <%= title %>" />\
+      <div class="filterMenu filterMenu-string">\
+        <div class="filter-text">Show:</div>\
+        <div class="icon-addon addon-sm">\
+          <input class="filter-string form-control input-sm" id="value" type="text" name="filter-string" />\
+          <label for="filter-string" class="glyphicon glyphicon-search"></label>\
+        </div>\
+        <button class="btn btn-sm btn-filter" name="button" type="submit" title="">Apply</button>\
+        <button class="btn btn-primary btn-sm btn-clear pull-right" name="button" type="submit" title="">Clear</button>\
```
nburwell: having these 2 buttons repeated for each filter is not DRY, can you make this part of the base class (or at worst case, a shared child component that is rendered in place for each filter type)

src/plugins/data_table/filter_view.js
```diff
@@ -68,43 +76,79 @@ var DataTableFilter = (function(options) {
         this.filter.value = filterInput.value;
         this.parentView._toggleIcon(true);
       }
+    },
+
+    afterRender: function() {
+      this.$('.btn-filter').click(this.parentView._onFilterClick.bind(this.parentView));
+      this.$('.btn-clear').click(this.parentView._onClearClick.bind(this.parentView));
```
nburwell: ditto on these afterRender's that are shared with all the filter views.  Esp since they have to call parentView.. seems this stuff should live up on the parent view!

src/plugins/data_table/filter_view.js
```diff
@@ -68,43 +76,79 @@ var DataTableFilter = (function(options) {
...
-        <li> = <input data-numeric-filter-name="eq" class="filter-numeric filter-numeric-equal" type="text" /></li> \
-      ', null, DEFAULT_JST_DELIMS),
+      <div class="filterMenu filterMenu-numeric">\
+        <div class="filter-text">Show items with value that:</div>\
+        <select class="filter-type form-control" data-filter-id="first-filter">\
+          <option selected value="gt">is greater than</option> \
+          <option value="lt">is less than</option> \
+          <option value="eq">is equal to</option> \
+        </select> \
+        <input id="first-filter" class="filter-value form-control" type="text" data-filter-type="gt"  /> \
```
nburwell: do you need the `id` attrs here and in the second filter? Hopefully not.. if so let's make them classes instead

src/plugins/data_table/filter_view.js
```diff
@@ -68,43 +76,79 @@ var DataTableFilter = (function(options) {
...
+        <button class="btn btn-primary btn-sm btn-clear pull-right" name="button" type="submit" title="">Clear</button>\
+      </div>\
+    ', null, DEFAULT_JST_DELIMS),
 
     _onInputChange: function (event) {
-      var filterInput = event.target;
-      var numericValueName = $(filterInput).attr("data-numeric-filter-name");
-      if (filterInput.value === "") {
-        this.filter[numericValueName] = null;
+      event.stopImmediatePropagation();
```
nburwell: why do we need to stop propagation?

src/plugins/data_table/filter_view.js
```diff
@@ -206,35 +257,23 @@ var DataTableFilter = (function(options) {
     },
 
     _onToggleClick: function (event) {
```
nburwell: do we still need this 1 line method that just stops the event?  maybe there's another solution .. do we need a listener here at all?

src/plugins/data_table/filter_view.js
```diff
@@ -206,35 +257,23 @@ var DataTableFilter = (function(options) {
...
+      this.$(".toggle-filter-button").popoverMenu('hide');
     },
 
     _onClearClick: function () {
-      $("input[type=text]", this.head).val("");
-      $("input[type=checkbox]", this.head).attr("checked", false);
-      $("input", this.head).trigger("change");
+      $("input[type=text]", this.child("filter-menu").$el).val("");
+      $("input[type=checkbox]", this.child("filter-menu").$el).attr("checked", false);
+      $("input", this.child("filter-menu").$el).trigger("change");
```
nburwell: hmm seems a better encapsulation would be for this base class to have the buttons rendered (as alluded to above in DRYing up the code), and keep these click handlers, but this onClearClick could emit an event that the child views listened for, and each was in charge of clearing their own data.  Rather than reaching into these filter-menus

src/plugins/data_table/filter_view.js
```diff
@@ -190,12 +238,15 @@ var DataTableFilter = (function(options) {
         filterButtonClass: this.filterButtonClass
       }));
 
-      $(".filterMenu", this.$el).prepend(this.child("filter-menu").render().$el);
+      this.$('.toggle-filter-button').popoverMenu({
+        content: this.child("filter-menu").render().$el
```
nburwell: right here.. I think I would render the filter menu, then also render a filter-actions template and append / concatenate together

mileszim: Addressed the rendering in the parent filter menu class.

src/plugins/data_table/filter_view.js
mileszim: If we don't stop propagation here, it triggers the sorting when you click the filter menu toggle button.

src/plugins/data_table/filter_view.js
mileszim: Made a remark on the propagation above as why this exists

Gruntfile.js
mileszim: The bootstrap is needed in the tests for the popoverMenu. I'm not sure what a better way would be. Any suggestions?

src/plugins/data_table/filter_view.js
mileszim: Ended up finding a use for this anyway!

src/plugins/data_table/filter_view.js
mileszim: Ended up finding a use for this anyway!

mileszim:
@nburwell backdraft changes are ready for second CR!

vendor/popover_menu.js
mileszim: I actually hadn't given that much thought, but I suppose that does make sense.

nburwell: hmm if we only need this for tests (in this repo), we could also just get a stub of it (and bootstrap) for now.  That may be the way to go here

Gruntfile.js
nburwell: If only needed for the specs, we could stub the smallest amount needed (vendor/mock-bootstrap.js) ?  But if we want examples to show the actual filtering working, which would be nice.. I don't have any suggestion for this pass

nburwell:
nice! :+1: unless we want to do anything about the BS.  But I'm fine deferring

PR #45, opened 2015-11-18
91 ux column resize bug: change .width to .outerWidth as it accounts …

albertovillalobos:
…for the whole computer width

@nburwell Hey Nick I found a bug with the resize that I just fixed.

To avoid shrinking adjacent columns when a column is being resized, we keep a "limit" by getting the starting width of the adjacent column so that it doesn't get smaller than what it is.

This was calculated with .width() and not .outerWidth(). So when x column was dragged into a smaller size, the one to it's left would shrink by roughly 10 pixels. This was caused because the limit was set to it's width without border or padding. With outerWidth() it accounts for those so that the adjacent columns don't shrink at all.

This bug only showed up if one expanded to adjacent columns and then shrunk the right column causing the left column to shrink by roughly 10 pixels (border+ padding)

### DIFF
--- a/dist/backdraft.js
+++ b/dist/backdraft.js
@@ -3174,24 +3174,23 @@ $.extend( $.fn.dataTableExt.oPagination, {
         this.s.mouse.startX = e.pageX;
         this.s.tableWidth = $(nTh).closest("table").width();
 
-
         // If we are at the left end, we expand the previous column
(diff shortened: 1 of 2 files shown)


nburwell:
good find. Hopefully works in all modes and not just based on the `box-sizing` setting

nburwell:
:+1:

PR #46, opened 2015-11-19
91 filter limit error message

mileszim:
Web changes: https://github.com/Invoca/web/pull/4333


mileszim:
@nburwell This is ready for a CR

nburwell:
:+1:

PR #47, opened 2015-11-21
phantomjs bind incompatibility fixes

mileszim:
@nburwell FYI

Tests pass, will merge if Invoca/web build is clean


nburwell:
nice thanks. This keeps our backdraft repo compatible with more browsers too =)

nburwell:
:+1:

PR #49, opened 2015-11-24
91 t backdraft export fix: add extra url param to specify backraft re…

albertovillalobos:
…quest

@nburwell backdraft side, cr plz


nburwell:
In the future the name `backdraft_request` (and whether to include it at all) should probably be configurable, since this is an open source repo and meant to be generic code, not coupled to our usage in our app.

And likewise that `ext_filter_json` param. Gross. That should be encapsulated and just made a configurable option too

nburwell:
But, for now.. :+1:

PR #50, opened 2015-11-30
Allow model removal from server side collections

SCommette:


nburwell:
Nice!  Looks good to me :+1:

PR #51, opened 2016-01-06
93 WEB-1830 column picker requests

mileszim:
https://ringrevenue.atlassian.net/browse/WEB-1830

Sibling Web PR: https://github.com/Invoca/web/pull/4585

### DIFF
--- a/dist/backdraft.js
+++ b/dist/backdraft.js
@@ -1290,14 +1290,18 @@ _.extend(Plugin.factory, {
       return this.dataTable.fnSettings().fnRecordsTotal();
     },
 
+    columnRequired: function(state, title) {
+      if (!state && this._columnManager.columnConfigForTitle(title).required) {
+        throw new Error("can not disable visibility when column is required");
+      }
+    },
+
     columnVisibility: function(title, state) {
       if (arguments.length === 1) {
         // getter
         return this._columnManager.visibility.get(title);
       } else {
-        if (!state && this._columnManager.columnConfigForTitle(title).required) {
-          throw new Error("can not disable visibility when column is required");
-        }
+        this.columnRequired(state, title);
         this._columnManager.visibility.set(title, state);
         state && this.renderColumn(title);
       }
@@ -1305,11 +1309,7 @@ _.extend(Plugin.factory, {
 
     // takes a hash of { columnTitle: columnState, ... }
     setColumnVisibilities: function(columns) {
-      _.each(columns, function(state, title) {
-        if (!state && this._columnManager.columnConfigForTitle(title).required) {
-          throw new Error("can not disable visibility when column is required");
-        }
-      }, this);
+      _.each(columns, this.columnRequired, this);
       this._columnManager.visibility.set(columns);
       _.each(columns, function(state, title) {
         state && this.renderColumn(title);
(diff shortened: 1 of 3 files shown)


mileszim:
@nburwell CR?

dist/backdraft.js
```diff
@@ -1303,6 +1303,19 @@ _.extend(Plugin.factory, {
...
     },
 
+    // takes a hash of { columnTitle: columnState, ... }
+    setColumnVisibilities: function(columns) {
+      _.each(columns, function(state, title) {
+        if (!state && this._columnManager.columnConfigForTitle(title).required) {
+          throw new Error("can not disable visibility when column is required");
+        }
+      }, this);
+      this._columnManager.visibility.set(columns);
```
nburwell: is this line critical to do in bulk like you are doing here?  Because otherwise, this whole method could become:

```
setColumnVisibilities: function(columns) {
  _.each(columns, function(state, title) {
    this.columnVisibility(title, state);
  }
}
```

That's way DRYer too..

mileszim: Bulk is exactly the whole reason I am doing this. Basically, when doing an _.each over the column visibility, every changed column triggers a "change" event, which triggers an ajax request for every column. When bulk changing, it triggers only one change event, and thus only one ajax request to the server.

dist/backdraft.js
```diff
@@ -1303,6 +1303,19 @@ _.extend(Plugin.factory, {
       }
     },
 
+    // takes a hash of { columnTitle: columnState, ... }
+    setColumnVisibilities: function(columns) {
+      _.each(columns, function(state, title) {
```
nburwell: so this is `state, title` but the method above (and lower level `columnManager.visibility()`) takes the arguments as `title, state`

It's too bad they get yielded as value, key here.. not sure there is much to be done here short of a manual loop construct instead of `_.each`

nburwell:
A note and a question is all!

dist/backdraft.js
nburwell: but there's also `changeColumnVisibilities` in web side, isn't that turning off the "saveOnVisibilityChanges", then invoking this code, then turning it back on?  And if not, what is `this._settingsManager.saveOnVisibilityChanges = saveVisibility || userInitiated;` line really doing..

mileszim: My understanding is saveOnVisibilityChanges is a flag that lets you decide to persist or not persist the visibility changes. In this line you can see it used https://github.com/Invoca/web/blob/93/t/WEB-1830_column_picker_requests/vendor/backdraft/assets/plugins/src/sparks/action_table.js#L783

So ```this._settingsManager.saveOnVisibilityChanges = saveVisibility || userInitiated;``` says "we need to persist if either specified app-level or specific at the time this function is called. If we not specified app-level or when calling the function, don't persist."

nburwell: You're right, k yeah that is another level of control.  And after your earlier comment I looked further into that `this._columnManager.visibility.set(columns)` interface (and how we also call `.set(title, state)` else where.. it's a confusing interface in that it accepts bulk or just one, and is okay with going from a big hash to just setting one (therefore blowing away the other values) as I think that Backbone model is only used transiently to fire change events ... so yeah, like you said this allows just one change event to fire versus 1 for each.. 

But it bugs me that the check & raise is the same as the method just above it, and this each loop down below is the same too

mileszim:
@nburwell some minor updates, ready for final CR if needed.

nburwell:
cool thanks! :+1:

PR #53, opened 2016-03-18
96/t/STORY-3818_urls_not_updating_w_filters

albertovillalobos:
WIP

### DIFF
--- a/dist/backdraft.js
+++ b/dist/backdraft.js
@@ -886,8 +886,9 @@ _.extend(Plugin.factory, {
       // get url parameters into an array
       var params=[];
       // if there are already parameters there, get them
-      if (window.location.href.split("?")[1]) {
-        params = $.deparam(window.location.href.split("?")[1]);
+      var urlArray = window.location.href.split("?");
+      if (urlArray[1]) {
+        params = $.deparam(urlArray[1]);
       }
       // get the filter settings
       var filteringSettings = this.parent.table._getFilteringSettings();
@@ -906,7 +907,7 @@ _.extend(Plugin.factory, {
       // if history is supported, add it to the url
       if (history.replaceState) {
         var state = { params: params };
-        var url =  window.location.href.split("?")[0] + "?" + jQuery.param(params);
+        var url =  urlArray[0] + "?" + jQuery.param(params);
         history.replaceState(state, window.document.title, url);
       }
     },
@@ -1237,7 +1238,8 @@ _.extend(Plugin.factory, {
       var params = $.deparam(splitUrl[1]);
 
       // make ext_filter_json param the same as the current url, now with new filters
-      params.ext_filter_json = $.deparam(window.location.href.split("?")[1]).ext_filter_json || $.deparam(window.location.href.split("?")[1]).filter_json;
+      var urlParamString = window.location.href.split("?")[1];
+      params.ext_filter_json = $.deparam(urlParamString).ext_filter_json || $.deparam(urlParamString).filter_json;
 
       // Build new url with old endpoint but new params
       var newURL = endpoint + "?"+ $.param(params);
--- a/dist/backdraft.js
+++ b/dist/backdraft.js
@@ -515,9 +515,16 @@ _.extend(Plugin.factory, {
     this._computeColumnLookups();
   },
 
-  _getUrlParams: function(url) {
-    return $.deparam(url);
+  _getUrlFilterParams: function() {
+    var urlParamString = window.location.href.split("?")[1];
+    if (urlParamString && $.deparam(urlParamString) && ($.deparam(urlParamString).filter_json || $.deparam(urlParamString).ext_filter_json) ) {
+      return JSON.parse($.deparam(urlParamString).filter_json || $.deparam(urlParamString).ext_filter_json);
+    }
+    else {
+      return []
+    }
   },
+
 
   _computeColumnConfig: function() {
     this.dataTableColumns = [];
@@ -536,19 +543,13 @@ _.extend(Plugin.factory, {
       return !columnConfig.bulk;
     });
 
-    // read initial filters from URL
-    var urlParamString = window.location.href.split("?")[1];
-    if (urlParamString && this._getUrlParams(urlParamString) && (this._getUrlParams(urlParamString).filter_json || this._getUrlParams(urlParamString).ext_filter_json) ) {
-      // if they don't come in the new filter_json, check for ext_filter_json
-      var filterArray = this._getUrlParams(urlParamString).filter_json || this._getUrlParams(urlParamString).ext_filter_json
-      var urlFilters = JSON.parse(filterArray);
-      urlFilters.forEach(function(element, index, array){
-        var columnConfigIndex = _.findIndex(this.columnsConfig, {attr: element.attr});
-        if (columnConfigIndex >= 0) {
-          this.columnsConfig[columnConfigIndex].filter[element.comparison] = element.value;
-        }
-      }.bind(this));
-    }
+    var urlFilters = this._getUrlFilterParams();
+    urlFilters.forEach(function(element, index, array){
+      var columnConfigIndex = _.findIndex(this.columnsConfig, {attr: element.attr});
+      if (columnConfigIndex >= 0) {
+        this.columnsConfig[columnConfigIndex].filter[element.comparison] = element.value;
+      }
+    }.bind(this));
 
     _.each(this._determineColumnTypes(), function(columnType, index) {
       var config = this.columnsConfig[index];
@@ -1242,8 +1243,7 @@ _.extend(Plugin.factory, {
       var params = $.deparam(splitUrl[1]);
 
       // make ext_filter_json param the same as the current url, now with new filters
-      var urlParamString = window.location.href.split("?")[1];
-      params.ext_filter_json = $.deparam(urlParamString).ext_filter_json || $.deparam(urlParamString).filter_json;
+      params.ext_filter_json = JSON.stringify(this.table._columnManager._configGenerator._getUrlFilterParams());
 
       // Build new url with old endpoint but new params
       var newURL = endpoint + "?"+ $.param(params);
--- a/dist/backdraft.js
+++ b/dist/backdraft.js
@@ -868,6 +868,28 @@ _.extend(Plugin.factory, {
       // to be implemented by subclasses
     },
 
+    _updateFilterUrlParams: function() {
+      // get url parameters into an array
+      var params = $.deparam(window.location.href.split("?")[1]);
+      // get the filter settings
+      var filteringSettings = this.parent.table._getFilteringSettings();
+
+      // if there are active filters, put them in the ext_filter_json param
+      if (JSON.parse(filteringSettings).length>0) {
+        params.ext_filter_json = filteringSettings;
+      }
+      // otherwise delete the ext_filter_json param to keep a clean uri
+      else {
+        delete params.ext_filter_json;
+      }
+      // if history is supported, add it to the url
+      if (history.pushState) {
+        var state = { params: params };
+        var url =  window.location.href.split("?")[0] + "?" + jQuery.param(params);
+        history.pushState(state, window.document.title, url);
+      }
+    },
+
     disableFilter: function(errorMessage) {
       if (!this.enabled) return;
       this.$('.filterMenu').prepend(this.errorTemplate({ errorCopy: errorMessage }));
@@ -903,6 +925,7 @@ _.extend(Plugin.factory, {
         this.filter.value = filterInput.value;
         this.parentView._toggleIcon(true);
       }
+      this._updateFilterUrlParams();
     },
 
     clear: function() {
@@ -941,6 +964,7 @@ _.extend(Plugin.factory, {
         this.filter[filterType] = filterValue;
         this.parentView._toggleIcon(true);
       }
+      this._updateFilterUrlParams();
     },
 
     afterRender: function() {
@@ -1015,6 +1039,7 @@ _.extend(Plugin.factory, {
           this.parentView._toggleIcon(false);
         }
       }
+      this._updateFilterUrlParams();
     },
 
     clear: function() {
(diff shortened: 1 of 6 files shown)


albertovillalobos:
@Vicentecarrillo @codyjca CR?
cc: @mikeweaver

dist/backdraft.js
```diff
@@ -532,6 +532,16 @@ _.extend(Plugin.factory, {
...
     });
 
+    // read initial filters from URL
+    var urlParamString = window.location.href.split("?")[1];
+    if (urlParamString && $.deparam(urlParamString) && $.deparam(urlParamString).ext_filter_json ) {
+      var urlFilters = JSON.parse($.deparam(urlParamString).ext_filter_json);
+      urlFilters.forEach(function(element, index, array){
+        var columnConfigIndex = _.findIndex(this.columnsConfig, {attr: element.attr});
+        this.columnsConfig[columnConfigIndex].filter[element.comparison] = element.value;
+      }.bind(this));
```
mikeweaver: What is the purpose of this bind?

dist/backdraft.js
```diff
@@ -532,6 +532,16 @@ _.extend(Plugin.factory, {
       return !columnConfig.bulk;
     });
 
+    // read initial filters from URL
+    var urlParamString = window.location.href.split("?")[1];
+    if (urlParamString && $.deparam(urlParamString) && $.deparam(urlParamString).ext_filter_json ) {
+      var urlFilters = JSON.parse($.deparam(urlParamString).ext_filter_json);
+      urlFilters.forEach(function(element, index, array){
+        var columnConfigIndex = _.findIndex(this.columnsConfig, {attr: element.attr});
+        this.columnsConfig[columnConfigIndex].filter[element.comparison] = element.value;
```
mikeweaver: What if the column is not found? Seems like this could happen.

dist/backdraft.js
```diff
@@ -868,6 +878,36 @@ _.extend(Plugin.factory, {
...
+    _updateFilterUrlParams: function() {
+      // get url parameters into an array
+      var params;
+      // if there are already parameters there, get them
+      if (window.location.href.split("?")[1]) {
+        params = $.deparam(window.location.href.split("?")[1]);
+      }
+      // if not, params will be empty
+      else {
+        params = []
```
mikeweaver: You could init params to [] and skip the else block.

dist/backdraft.js
```diff
@@ -532,6 +532,16 @@ _.extend(Plugin.factory, {
       return !columnConfig.bulk;
     });
 
+    // read initial filters from URL
+    var urlParamString = window.location.href.split("?")[1];
+    if (urlParamString && $.deparam(urlParamString) && $.deparam(urlParamString).ext_filter_json ) {
```
mikeweaver: I thought we were going to rename the ext_filter_json param so it wasn't called EXT anymore. This was to prevent future developer confusion.

dist/backdraft.js
```diff
@@ -894,6 +934,20 @@ _.extend(Plugin.factory, {
       </div>\
       ', null, DEFAULT_JST_DELIMS),
 
+      afterRender: function() {
+        var filterArray = JSON.parse(this.parent.table._getFilteringSettings()) || [];
+        // if there are filters in the url...
+        if (filterArray.length > 0) {
+          // find the filters that match this filter instance
+          var matches = _.where(filterArray, {type: "string", attr: this.attr});
+          // if there are url params for this filter...
```
mikeweaver: What does the ... mean? I think you mean is this: // if there are url params for this filter then enable those filters in the UI

dist/backdraft.js
mikeweaver: Nevermind, figured it out.

dist/backdraft.js
```diff
@@ -1105,16 +1205,43 @@ _.extend(Plugin.factory, {
...
+      this._updateAjaxSource();
       this.table.dataTable._fnAjaxUpdate();
       this.$(".toggle-filter-button").popoverMenu('hide');
     },
 
+    _updateAjaxSource: function() {
+      // get ajax url
+      var ajaxURL = this.parent.dataTable.fnSettings().sAjaxSource;
+      // get the endpoint of ajax url
+      var endpoint = ajaxURL.split("?")[0];
```
mikeweaver: You split this three times in a row, perhaps saving the split in a local would be a good idea.

spec/plugins/data_table/server_side.js
```diff
@@ -775,12 +795,221 @@ describe("DataTable Plugin", function() {
...
+      cg._computeColumnConfig();
+      // get new filtering settings
+      var newFilterSettings = table._getFilteringSettings();
+      // They should be the same
+      expect(oldFilterSettings).toEqual(newFilterSettings);
+
+      history.pushState(null, null, "_SpecRunner.html");
+
+      // remove filters again so that other tests don't fail
+      table.dataTable.find("thead th").each(function (index) {
```
mikeweaver: Is there a common teardown function that you can move this into?

spec/plugins/data_table/server_side.js
```diff
@@ -775,12 +795,221 @@ describe("DataTable Plugin", function() {
...
+      var cg = table.configGenerator();
+      var expectedFilters = [];
+
+      // populate url with filters
+      table.dataTable.find("thead th").each(function (index) {
+        var wrapper = $(".DataTables_sort_wrapper", this);
+        if (wrapper) {
+          var title = wrapper.text();
+          var col = cg.columnConfigByTitle.attributes[title];
+          if (col && col.filter) {
```
mikeweaver: There is a bit of repetition in these tests. Can you DRY them up?

dist/backdraft.js
albertovillalobos: Addressed

dist/backdraft.js
albertovillalobos: addressed

dist/backdraft.js
albertovillalobos: I thought we decided against it to keep links compatible without extra logic.

codyjca: I agree we discussed renaming ext_filter_json but also I think we talked about still accepting ext_filter_json to make sure we support bookmarked links.

dist/backdraft.js
albertovillalobos: addressed

dist/backdraft.js
albertovillalobos: I would suggest we create a small tech item for the future, once we don't use old reports, and completely replace ext_filter_json with a small logic for convert `ext_filter_json` to `filter_json`? It would be a half day item and that way this story makes it into QA by today

dist/backdraft.js
albertovillalobos: addressed

spec/plugins/data_table/server_side.js
albertovillalobos: addressed

dist/backdraft.js
Vicentecarrillo: I would vote for implementing it while we are here, seems like QA has a bit of work to get through at the moment so I think that gives us some time. If we end up waiting on this story at any point then that might be a good idea

spec/plugins/data_table/server_side.js
```diff
@@ -673,13 +673,79 @@ describe("DataTable Plugin", function() {
...
 
+    function clearFilters() {
+      var cg = table.configGenerator();
+      table.dataTable.find("thead th").each(function (index) {
+        var wrapper = $(".DataTables_sort_wrapper", this);
+        if (wrapper) {
+          var title = wrapper.text();
+          var col = cg.columnConfigByTitle.attributes[title];
+          if (col && col.filter) {
+            if (col.filter.type === "string") {
```
Vicentecarrillo: probably prefer to use a switch statement here, ie:
```
switch (expression) {
  case value1:
    //Statements executed when the result of expression matches value1
    [break;]
  case value2:
    //Statements executed when the result of expression matches value2
    [break;]
  ...
  default:
    //Statements executed when none of the values match the value of the expression
    [break;]
}
```

spec/plugins/data_table/server_side.js
```diff
@@ -673,13 +673,79 @@ describe("DataTable Plugin", function() {
...
+    };
+
+    function populateFilters() {
+      var cg = table.configGenerator();
+      table.dataTable.find("thead th").each(function (index) {
+        var wrapper = $(".DataTables_sort_wrapper", this);
+        if (wrapper) {
+          var title = wrapper.text();
+          var col = cg.columnConfigByTitle.attributes[title];
+          if (col && col.filter) {
```
Vicentecarrillo: would `if (col.filter)` have the same logic as `if (col && col.filter)` ?

spec/plugins/data_table/server_side.js
```diff
@@ -673,13 +673,79 @@ describe("DataTable Plugin", function() {
...
+            if (col.filter.type === "string") {
```
Vicentecarrillo: switch statement here

dist/backdraft.js
mikeweaver: +1 on doing it today.

spec/plugins/data_table/server_side.js
```diff
@@ -673,13 +673,79 @@ describe("DataTable Plugin", function() {
...
     }
+    function verifyUrlParams(filterObj) {
+      var uri = encodeURIComponent($.deparam(window.location.href.split("?")[1]).ext_filter_json)
+      var expectedFilterJson
+      if (filterObj.length>0) {
+        expectedFilterJson = encodeURIComponent(JSON.stringify(filterObj));
+      }
+      else {
+        expectedFilterJson = "";
+      }
```
Vicentecarrillo: i think this would be a good use of a ternary
`var expectedFilterJson = (filterObj.length>0) ? encodeURIComponent(JSON.stringify(filterObj)) : expectedFilterJson = "";`

spec/plugins/data_table/server_side.js
mikeweaver: +1 on the switch statement

spec/plugins/data_table/server_side.js
```diff
@@ -775,12 +849,86 @@ describe("DataTable Plugin", function() {
...
+      cg.table.children["filter-type"].children["filter-menu"].afterRender()
+
+      // Check the filters are there
+      table.dataTable.find("thead th").each(function (index) {
+        var wrapper = $(".DataTables_sort_wrapper", this);
+        if (wrapper) {
+          var title = wrapper.text();
+          var col = cg.columnConfigByTitle.attributes[title];
+          if (col && col.filter) {
+            if (col.filter.type === "string") {
```
Vicentecarrillo: same comments

spec/plugins/data_table/server_side.js
mikeweaver: Nice, looks good.

src/plugins/data_table/filter_view.js
```diff
@@ -85,6 +85,32 @@ var DataTableFilter = (function(options) {
...
+      }
+      // otherwise delete the ext_filter_json param to keep a clean uri
+      else {
+        delete params.ext_filter_json;
+      }
+      // if history is supported, add it to the url
+      if (history.pushState) {
+        var state = { params: params };
+        var url =  window.location.href.split("?")[0] + "?" + jQuery.param(params);
+        history.pushState(state, window.document.title, url);
```
Vicentecarrillo: do we want to go with pushState over replaceState? might be something to UX's input on

albertovillalobos: Using pushstate would allow us to to hit the back button to previous configs.

spec/plugins/data_table/server_side.js
albertovillalobos: We don't use ternary as per the style guide. @nburwell what do you think?

src/plugins/data_table/filter_view.js
Vicentecarrillo: i know but is that what we want?

spec/plugins/data_table/server_side.js
Vicentecarrillo: hmm i just checked and didn't see it in there... i'll defer to @nburwell

nburwell: We're not against it. Maybe we should add a note saying we are fine with it?

The only mention of ternary was on simplying this case `var minCount = count ? count : 0; ` down to the more concise and DRYer `var minCount = count || 0;`

src/plugins/data_table/filter_view.js
nburwell: Thanks for bringing up Vicente. For things that change as often as filters do.. I think replaceState makes more sense?  No one wants to have to hit back 6 times to get to summary view if they played with the filters a bit (IMO).

Can we defer to what we decided and has seemed to work well for a few years on the existing reports for filter changes?  Please check and confirm it matches behavior there

Vicentecarrillo: i checked current detail reports and they don't use the history api for filter changes

albertovillalobos: Old reports do use the history API.

dist/backdraft.js
albertovillalobos: Sounds good, I'll work on this.

spec/plugins/data_table/server_side.js
albertovillalobos: Addressed

spec/plugins/data_table/server_side.js
albertovillalobos: Addressed

spec/plugins/data_table/server_side.js
albertovillalobos: Addressed

spec/plugins/data_table/server_side.js
albertovillalobos: Taking your suggestion @Vicentecarrillo

spec/plugins/data_table/server_side.js
albertovillalobos: Addressed

src/plugins/data_table/filter_view.js
nburwell: yes, they definitely do.. @albertovillalobos are they pushing a new history entry when filters change, or replacing current entry?  Please confirm here on this thread and update to match that behavior if necessary

albertovillalobos: @nburwell They use `replaceState`. I will be using that.

Vicentecarrillo: cool, yea sorry I meant to say they don't use pushState from checking the app

albertovillalobos:
@mikeweaver @Vicentecarrillo @codingiam I've addressed the comment regarding the ext_filter_json param

dist/backdraft.js
```diff
@@ -532,6 +532,20 @@ _.extend(Plugin.factory, {
       return !columnConfig.bulk;
     });
 
+    // read initial filters from URL
+    var urlParamString = window.location.href.split("?")[1];
+    if (urlParamString && $.deparam(urlParamString) && ($.deparam(urlParamString).filter_json || $.deparam(urlParamString).ext_filter_json) ) {
```
mikeweaver: There are three places in this file where the fetching of the filter params is repeated. Can you dry that up? (It will probably read easier that way too.)

albertovillalobos: Addressed

mikeweaver: Did you push it? Don't see it.

albertovillalobos: I think so yeah, unless you meant something else?

albertovillalobos: Addressed

dist/backdraft.js
```diff
@@ -532,6 +543,14 @@ _.extend(Plugin.factory, {
       return !columnConfig.bulk;
     });
 
+    var urlFilters = this._getUrlFilterParams();
```
mikeweaver: Why have a local var here?

albertovillalobos: Without the the next line would look like 
```
this._getUrlFilterParams().forEach(function(element, index, array){
```

mikeweaver:
:+1:

PR #55, opened 2016-03-30
96/t/WEB-1997: Fix comparison/numeric filters

ibiziiac-sv:
@codyjca please review. I have ported the changes and the spec, also made some adjustments in the `afterRender` method.

### DIFF
--- a/src/plugins/data_table/filter_view.js
+++ b/src/plugins/data_table/filter_view.js
@@ -213,12 +213,13 @@ var DataTableFilter = (function(options) {
 
     _onSelectChange: function (event) {
       event.stopImmediatePropagation();
-      var filterElementId = event.target.getAttribute('data-filter-id'),
-          previousFilterType = event.target.getAttribute('data-previous-value'),
-          filterType = event.target.value;
+      var target = $(event.target),
+          filterElementId = target.data('filter-id'),
+          previousFilterType = target.data('previous-value'),
+          filterType = target.val();
       this.filter[filterType] = this.filter[previousFilterType];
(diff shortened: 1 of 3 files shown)


codyjca:
:+1: thanks Ion!

Vicentecarrillo:
@nburwell do these changes to the datatable filters look safe enough to merge into the release branch before the deploy?
cc: @mikeweaver @deeptidjjain @sherriman

nburwell:
This seems safe to me, especially since the only consumer of the filter code here is the reporting app in web (so, isolated), and it is controlled by a feature flag if issues arise.

I'm assuming it's also been or will be QA'ed before merging too..

src/plugins/data_table/filter_view.js
```diff
@@ -206,6 +211,18 @@ var DataTableFilter = (function(options) {
       this._updateFilterUrlParams();
     },
 
+    _onSelectChange: function (event) {
+      event.stopImmediatePropagation();
+      var filterElementId = event.target.getAttribute('data-filter-id'),
+          previousFilterType = event.target.getAttribute('data-previous-value'),
+          filterType = event.target.value;
```
nburwell: why not use jQuery for these?

```
var target = $(event.target),
    filterElementId = target.data('filter-id'),
    previousFilterType = target.data('previous-value'),
    filterType = target.val();

...
target.data('previous-value', filterType);
```

Vicentecarrillo:
cool, that was my impression as well. It's being reQA'd at the moment. thanks!

PR #56, opened 2016-03-30
hotfix/t/WEB-2012_new_detail_transactions_friendly_name_bug: stop usi…

albertovillalobos:
…ng title to match html nodes to configs

Here's the bug fix.

@codyjca @Vicentecarrillo 
cc @mikeweaver @nburwell

### DIFF
--- a/dist/backdraft.js
+++ b/dist/backdraft.js
@@ -1275,15 +1275,14 @@ _.extend(Plugin.factory, {
 
     render : function() {
       var cells = this.findCells(), node;
-      var configsArray = this.columnsConfig;
-      var cleanConfigs = this.columnsConfig.map(function(a,b,c){
-        if (b===0) {
-          return a
+      var cleanConfigs = this.columnsConfig.map(function(currentValue,index){
+        if (index===0) {
+          return currentValue
         }
-        if(a.attr&&a.visible){
-          return a}
-        }).filter(function(n){
-           return n != undefined
+        if(currentValue.attr&&currentValue.visible){
+          return currentValue}
+        }).filter(function(currentValue){
+           return currentValue != undefined
          }
        );
       _.each(cleanConfigs, function(config, index) {
--- a/dist/backdraft.js
+++ b/dist/backdraft.js
@@ -1275,8 +1275,19 @@ _.extend(Plugin.factory, {
 
     render : function() {
       var cells = this.findCells(), node;
-      _.each(this.columnsConfig, function(config) {
-        node = cells.filter(config.nodeMatcher(config));
+      var configsArray = this.columnsConfig;
+      var cleanConfigs = this.columnsConfig.map(function(a,b,c){
+        if (b===0) {
+          return a
+        }
+        if(a.attr&&a.visible){
+          return a}
+        }).filter(function(n){
+           return n != undefined
+         }
+       );
+      _.each(cleanConfigs, function(config, index) {
+        node = $(cells[index]);
         this._invokeRenderer(config, node);
       }, this);
     },
@@ -1325,6 +1336,7 @@ _.extend(Plugin.factory, {
   return Row;
 
 })();
+
   var LocalDataTable = (function() {
 
   var Base = Backdraft.plugin("Base");
@@ -1763,13 +1775,13 @@ _.extend(Plugin.factory, {
         if (col) {
           // We only make the filter controls if there's a filter element in the column manager
           if (col.filter) {
-            table.child("filter-"+col.attr, new DataTableFilter({
+            table.child("filter-"+cg.columnsConfig[index].attr, new DataTableFilter({
               column: col,
               table: table,
               head: this,
               className: "dropdown DataTables_filter_wrapper"
             }));
-            $(this).append(table.child("filter-"+col.attr).render().$el);
+            $(this).append(table.child("filter-"+cg.columnsConfig[index].attr).render().$el);
           }
         }
       });
--- a/dist/backdraft.js
+++ b/dist/backdraft.js
@@ -76,9 +76,10 @@
 Backdraft.Utils.toCSSClass = (function() {
   var cssClass = /[^a-zA-Z_0-9\-]/g;
   return function(input) {
-    return input.replace(cssClass, "-");
+    return "column-"+input.replace(cssClass, "-");
   };
 })();
+
 
   var App = (function() {
 
@@ -1275,16 +1276,8 @@ _.extend(Plugin.factory, {
 
     render : function() {
       var cells = this.findCells(), node;
-      var cleanConfigs = this.columnsConfig.map(function(currentValue,index){
-        if (currentValue.bulk || ((currentValue.attr || currentValue.title) && currentValue.visible) ) {
-          return currentValue;
-        }
-      }).filter(function(currentValue) {
-         return currentValue !== undefined;
-        }
-      );
-      _.each(cleanConfigs, function(config, index) {
-        node = $(cells[index]);
+      _.each(this.columnsConfig, function(config) {
+        node = cells.filter(config.nodeMatcher(config));
         this._invokeRenderer(config, node);
       }, this);
     },
@@ -1769,16 +1762,26 @@ _.extend(Plugin.factory, {
         var title = this.textContent || this.innerText;
         var col = cg.columnConfigByTitle.attributes[title];
 
+        // Try to grab it by attribute, if possible
+        var matches = this.className.match(/(?:^|\s)column-(.*?)(?:\s|$)/);
+        if (matches && matches[1]) {
+          cg.columnsConfig.forEach(function(currentObject){
+            if (currentObject.attr && Backdraft.Utils.toCSSClass(currentObject.attr).replace("column-","") === matches[1]) {
+              col = currentObject;
+            }
+          })
+        }
+
         if (col) {
           // We only make the filter controls if there's a filter element in the column manager
           if (col.filter) {
-            table.child("filter-"+cg.columnsConfig[index].attr, new DataTableFilter({
+            table.child("filter-"+col.attr, new DataTableFilter({
               column: col,
               table: table,
               head: this,
               className: "dropdown DataTables_filter_wrapper"
             }));
-            $(this).append(table.child("filter-"+cg.columnsConfig[index].attr).render().$el);
+            $(this).append(table.child("filter-"+col.attr).render().$el);
           }
         }
       });
@@ -4194,7 +4197,7 @@ $.extend( $.fn.dataTableExt.oPagination, {
   });
 
   columnType.nodeMatcher(function(config) {
-    return "." + Backdraft.Utils.toCSSClass(config.title);
+    return "." + Backdraft.Utils.toCSSClass(config.attr || config.title);
   });
 
   columnType.definition(function(dataTable, config) {
@@ -4203,7 +4206,7 @@ $.extend( $.fn.dataTableExt.oPagination, {
       bSearchable: config.search,
       asSorting: config.sortDir,
       sTitle: config.title,
-      sClass : Backdraft.Utils.toCSSClass(config.title),
+      sClass : Backdraft.Utils.toCSSClass(config.attr || config.title),
       mData: function(source, type, val) {
         return dataTable.collection.get(source).get(config.attr);
       },
@@ -4235,7 +4238,7 @@ $.extend( $.fn.dataTableExt.oPagination, {
   });
 
   columnType.nodeMatcher(function(config) {
-    return "." + Backdraft.Utils.toCSSClass(config.title);
+    return "." + Backdraft.Utils.toCSSClass(config.attr || config.title);
   });
 
   columnType.definition(function(dataTable, config) {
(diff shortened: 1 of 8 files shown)


dist/backdraft.js
```diff
@@ -1275,8 +1275,19 @@ _.extend(Plugin.factory, {
 
     render : function() {
       var cells = this.findCells(), node;
-      _.each(this.columnsConfig, function(config) {
-        node = cells.filter(config.nodeMatcher(config));
+      var configsArray = this.columnsConfig;
+      var cleanConfigs = this.columnsConfig.map(function(a,b,c){
```
Vicentecarrillo: what the the parameters a,b,c? can you rename them to something more descriptive?

spec/plugins/data_table/server_side.js
```diff
@@ -747,6 +747,67 @@ describe("DataTable Plugin", function() {
...
+
+      jasmine.Ajax.install();
+      mockResponse = new MockResponse();
+
+      table = new app.Views.T({ collection : collection });
+      table.render();
+      jasmine.Ajax.requests.mostRecent().response(mockResponse.get());
+      history.pushState(null, null, "_SpecRunner.html");
+      cg = table.configGenerator();
+    });
```
Vicentecarrillo: not seeing any assertions in here?

dist/backdraft.js
albertovillalobos: Addressed

spec/plugins/data_table/server_side.js
albertovillalobos: I wasn't sure what to assert here. Basically without the fix the test fails because an error is thrown.

nburwell: I'd say to assert you have 2 filter views, one for cost and one for cost2 .. either soething that gets added to the dom, or that exists off of `cg`, ..

spec/plugins/data_table/server_side.js
```diff
@@ -747,6 +747,68 @@ describe("DataTable Plugin", function() {
...
+
+      collection = new app.Collections.Col();
+
+      jasmine.Ajax.install();
+      mockResponse = new MockResponse();
+
+      table = new app.Views.T({ collection : collection });
+      table.render();
+      jasmine.Ajax.requests.mostRecent().response(mockResponse.get());
+      expect(table).toBeDefined();
```
Vicentecarrillo: does this assertion guarantee that there aren't duplicate filter views?

albertovillalobos: Yes, the test will fail when there are duplicate filters table will not be defined. On top of that the test will error out because of a duplicate filter view.

dist/backdraft.js
```diff
@@ -1275,8 +1275,19 @@ _.extend(Plugin.factory, {
...
+        if (b===0) {
+          return a
+        }
+        if(a.attr&&a.visible){
```
nburwell: please match the styleguide for spacing around `if` and `{` etc, throughout this PR:

```
if (a.attr && a.visible) {
  ^       ^  ^          ^
```

dist/backdraft.js
```diff
@@ -1275,8 +1275,19 @@ _.extend(Plugin.factory, {
 
     render : function() {
       var cells = this.findCells(), node;
-      _.each(this.columnsConfig, function(config) {
-        node = cells.filter(config.nodeMatcher(config));
+      var configsArray = this.columnsConfig;
+      var cleanConfigs = this.columnsConfig.map(function(a,b,c){
+        if (b===0) {
+          return a
```
nburwell: Please throw a `;` at end of all the returns

dist/backdraft.js
```diff
@@ -1275,8 +1275,19 @@ _.extend(Plugin.factory, {
...
+        }
+        if(a.attr&&a.visible){
+          return a}
+        }).filter(function(n){
+           return n != undefined
```
nburwell: surprised jslint didn't squawk (do we not have it on backdraft?) but this should be `!==`.. or do we want `typeof n !== "undefined"` ?

dist/backdraft.js
```diff
@@ -1275,8 +1275,19 @@ _.extend(Plugin.factory, {
 
     render : function() {
       var cells = this.findCells(), node;
-      _.each(this.columnsConfig, function(config) {
-        node = cells.filter(config.nodeMatcher(config));
+      var configsArray = this.columnsConfig;
```
nburwell: don't see this used? please remove

dist/backdraft.js
```diff
@@ -1275,8 +1275,18 @@ _.extend(Plugin.factory, {
 
     render : function() {
       var cells = this.findCells(), node;
-      _.each(this.columnsConfig, function(config) {
-        node = cells.filter(config.nodeMatcher(config));
+      var cleanConfigs = this.columnsConfig.map(function(currentValue,index){
```
nburwell: What does `cleanConfigs` mean?  what are you cleaning up or normalizing? Can you describe it in the variable name so the code is more clear?

dist/backdraft.js
```diff
@@ -1275,8 +1275,18 @@ _.extend(Plugin.factory, {
...
+        if (index===0) {
```
nburwell: why does the first item have special treatment?

dist/backdraft.js
```diff
@@ -1275,8 +1275,18 @@ _.extend(Plugin.factory, {
...
+          return currentValue
+        }
+        if(currentValue.attr&&currentValue.visible){
```
nburwell: is this a net new change or no.. just trying to understand the implementation based on the code that is here... looks like now columns without an `attr` are skipped over, is that true? 

I think we have some tables that have columns with an `attr` key?  If my above assumption is true (and it may not be), why wouldn't we enforce `attr` as a required key up in the column config definition

dist/backdraft.js
albertovillalobos: cleanConfigs is the array of column configs without any non-visible or undefined configs. I can't come up with a better name, `definedConfigs`, `availableConfigs` ?

dist/backdraft.js
albertovillalobos: All columns should have attr except for the bulk column. So on this case I check that either they're bulk or have attr and are visible.

dist/backdraft.js
albertovillalobos: It was meant for the bulk column but now I should check it it's bulk not if it's index 0.

dist/backdraft.js
albertovillalobos: addressed

spec/plugins/data_table/server_side.js
albertovillalobos: Addressed

dist/backdraft.js
albertovillalobos: Addressed

dist/backdraft.js
albertovillalobos: Addressed

dist/backdraft.js
albertovillalobos: Addressed

albertovillalobos:
I've addressed all the CR, round 2?
@Vicentecarrillo 
cc @nburwell @mikeweaver

src/plugins/data_table/row.js
```diff
@@ -10,8 +10,16 @@ var Row = (function() {
...
       var cells = this.findCells(), node;
-      _.each(this.columnsConfig, function(config) {
-        node = cells.filter(config.nodeMatcher(config));
+      var cleanConfigs = this.columnsConfig.map(function(currentValue,index){
+        if (currentValue.bulk || ((currentValue.attr || currentValue.title) && currentValue.visible) ) {
+          return currentValue;
+        }
+      }).filter(function(currentValue) {
+         return currentValue !== undefined;
+        }
```
Vicentecarrillo: is this filter necessary? not following how an undefined config would get passed back from the map function

albertovillalobos: Yes, the map only replaces stuff, doesn't splice. So it'd look like `[object, object, undefined, object, object, undefined]`, the filter removes the undefined parts.

dist/backdraft.js
nburwell: But what if the first column is not bulk, but doesn't have an attr? You're blindly including the first column always.  Also the bulk column can be anywhere, not just the first position.. 

> All columns should have attr or title

Okay, that sounds reasonable, yes.. where is this code checking title though?

src/plugins/data_table/row.js
mikeweaver: Is that true? I thought map ignored undefined results?

dist/backdraft.js
mikeweaver: How about visibleColumnConfigs? I think that matches up with what you are doing here, trying to get the indexes of the visible columns.

Also, seems like this could be extracted into a function to make the code easier to read.

albertovillalobos:
Alright, things changed a lot now. 
@Vicentecarrillo @codyjca CR please
cc @nburwell @mikeweaver

src/plugins/data_table/data_table.js
```diff
@@ -433,6 +433,16 @@ var LocalDataTable = (function() {
         var title = this.textContent || this.innerText;
         var col = cg.columnConfigByTitle.attributes[title];
 
+        // Try to grab it by attribute, if possible
+        var matches = this.className.match(/(?:^|\s)column-(.*?)(?:\s|$)/);
```
mikeweaver: You could use column-([^\s]*) to make the regex a little safer. (It will not consume any spaces that way)

src/plugins/data_table/column_types/attr.js
```diff
@@ -4,7 +4,7 @@ app.view.dataTable.columnType(function(columnType) {
   });
 
   columnType.nodeMatcher(function(config) {
-    return "." + Backdraft.Utils.toCSSClass(config.title);
+    return "." + Backdraft.Utils.toCSSClass(config.attr || config.title);
```
mikeweaver: Why are we supporting both ways? Why not just attr?

albertovillalobos: Some backdraft tables don't use config.attr at all.

nburwell: ah, Mike's comment is a good one here though, I led you slightly astray yesterday Alberto.  Yes, definitely true that some tables don't use config.attr at all, and that is a main reason for the redo of the implementation here.. but this `column_types/attr.js` should only be used on columns where there **is** an attr. (See lines 2~4).

So in this file, I think it is correct / more accurate to just use config.attr on this line and line 16 below.

src/plugins/data_table/column_types/base.js
```diff
@@ -4,7 +4,7 @@ app.view.dataTable.columnType(function(columnType) {
   });
 
   columnType.nodeMatcher(function(config) {
-    return "." + Backdraft.Utils.toCSSClass(config.title);
+    return "." + Backdraft.Utils.toCSSClass(config.attr || config.title);
```
nburwell: and this file can just stay as config.title because notice it's only used when `!config.attr` (per line 3)

src/plugins/data_table/column_types/base.js
```diff
@@ -17,7 +17,7 @@ app.view.dataTable.columnType(function(columnType) {
       bSortable: sortable,
       bSearchable: searchable,
       sTitle: config.title,
-      sClass : Backdraft.Utils.toCSSClass(config.title),
+      sClass : Backdraft.Utils.toCSSClass(config.attr || config.title),
```
nburwell: ditto here

src/plugins/data_table/data_table.js
```diff
@@ -433,6 +433,16 @@ var LocalDataTable = (function() {
...
+        if (matches && matches[1]) {
+          cg.columnsConfig.forEach(function(currentObject){
+            if (currentObject.attr && Backdraft.Utils.toCSSClass(currentObject.attr).replace("column-","") === matches[1]) {
+              col = currentObject;
+            }
+          })
+        }
```
nburwell: wouldn't it be easier to follow the logic if the below line was the `else` block here? (and not always executed on L434 since we may just overwrite it..)

```
col = cg.columnConfigByTitle.attributes[title];
```

src/plugins/data_table/column_types/attr.js
albertovillalobos: I see, will do.

src/plugins/data_table/column_types/base.js
albertovillalobos: Got it

src/plugins/data_table/data_table.js
albertovillalobos: On it

Vicentecarrillo:
closing this and opening PR to backdraft integration branch

src/plugins/data_table/data_table.js
albertovillalobos: Addressed

src/plugins/data_table/column_types/base.js
albertovillalobos: Addressed

src/plugins/data_table/column_types/base.js
albertovillalobos: Addressed

src/plugins/data_table/column_types/attr.js
albertovillalobos: Addressed

PR #65, opened 2016-06-24
fixed bug around disabling column sort while in edit mode

dangrozasv:
@nburwell Please review.


nburwell:
as long as you have confirmed this doesn't mess up any header layouts (specifically in reports that has name, sort and filter icons), then 👍

PR #66, opened 2016-07-18
Fix bulk select delete

euge:
Fix bug where removing a model from the collection does not adjust the bulk selection of the model.


nburwell:
Thanks @euge , you rock!

PR #68, opened 2016-12-14
1611/t/STORY-4471.STORY-4679_datatable-column-management - updated to…

codyjca:
Updates to backdraft to support dynamically building the header group (used for comparison time periods report)

- Added header group example to the example server
- updated LocalDataTable to accept a new row config, columnGroupDefinitions. Much like the grand total row, if the data is passed in we will automatically build the header group.
-- For now we disabled reordering the columns when in this mode.

@nburwell CR? cc @mikeweaver 

image

### DIFF
--- a/dist/backdraft.js
+++ b/dist/backdraft.js
@@ -1627,34 +1627,63 @@ _.extend(Plugin.factory, {
       }
     },
 
+    _renderHeaderGroup: function() {
+      if (this.rowClass.prototype.columnGroupDefinitions) {
+        var columnGroups = this.rowClass.prototype.columnGroupDefinitions;
+        var tr = this.$("table").find('thead tr.header-groups-row');
+        if (tr.length === 0) {
+          tr = $('<tr class="header-groups-row">');
+        } else {
+          tr.empty();
+        }
+
+        var uniqueHeaderGroupDataIndex = {};
+
+        _.each(this._columnManager._visibilitySummary().visible, function(col) {
+          var columnConfig = _.findWhere(this._columnManager.columnsConfig(), { attr: col });
+          var headerGroupDataIndex = columnConfig.headerGroupDataIndex;
+          var columnGroupConfig = _.findWhere(columnGroups, { "headerGroupDataIndex" : headerGroupDataIndex } );
+
+          if (columnGroupConfig && !uniqueHeaderGroupDataIndex[headerGroupDataIndex]) {
+            uniqueHeaderGroupDataIndex[headerGroupDataIndex] = true;
+            tr.append('<th colspan="' + columnGroupConfig.colspan + '" rowspan="1" class="header-groups">' + columnGroupConfig.headerName + '</th>');
+          }
+        }.bind(this));
+
+        this.$("table").find('thead').prepend(tr);
+      }
+    },
+
     // Changes or resets the column order.
     // When called with no args, returns the current order.
     // Call with { reset : true } to have it restore column order to initial configuration
     // Provide array of indexes as first argument to have it reordered by that
     _changeColumnOrder: function(order) {
-      var columnsOrig = _.clone(this.dataTable.fnSettings().aoColumns);
-      if (_.isArray(order)) {
-        this._colReorder.fnOrder(order);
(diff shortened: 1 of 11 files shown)


spec/plugins/data_table/local.js
```diff
@@ -452,4 +452,72 @@ describe("DataTable Plugin", function() {
...
+        expect(table.$(".header-groups-row").text()).toEqual("User Info");
+        expect(table.$(".header-groups-row").children()[1].attributes.colspan.value).toEqual("2");
+        expect(table._colReorder).toEqual(undefined);
+      });
+
+      it("should disable column sorting (_colReorder) when using a header group", function() {
+        expect(table.$(".header-groups-row").length).toEqual(1);
+        expect(table._colReorder).toEqual(undefined);
+      });
+    });
```
nburwell: nice test! Can you add another test describing what happens if you supply `columnGroupDefinitions` but `headerGroupDataIndex` either is not defined on each column, or you don't find any columns of that headerGroupDataIndex.  Do you silently skip over them? is the header messed up? Should we raise an error?

src/plugins/data_table/data_table.js
```diff
@@ -277,34 +277,63 @@ var LocalDataTable = (function() {
...
+        var uniqueHeaderGroupDataIndex = {};
+
+        _.each(this._columnManager._visibilitySummary().visible, function(col) {
+          var columnConfig = _.findWhere(this._columnManager.columnsConfig(), { attr: col });
+          var headerGroupDataIndex = columnConfig.headerGroupDataIndex;
+          var columnGroupConfig = _.findWhere(columnGroups, { "headerGroupDataIndex" : headerGroupDataIndex } );
+
+          if (columnGroupConfig && !uniqueHeaderGroupDataIndex[headerGroupDataIndex]) {
+            uniqueHeaderGroupDataIndex[headerGroupDataIndex] = true;
+            tr.append('<th colspan="' + columnGroupConfig.colspan + '" rowspan="1" class="header-groups">' + columnGroupConfig.headerName + '</th>');
```
nburwell: minor nit.. you can omit `rowspan="1"` as that is the default in HTML

src/plugins/data_table/data_table.js
```diff
@@ -540,11 +573,13 @@ var LocalDataTable = (function() {
     _onDraw : function() {
       this.trigger("draw", arguments);
       this._renderGrandTotalsRow();
+      this._renderHeaderGroup();
     },
 
     _onColumnVisibilityChange: function(summary) {
       this.dataTable.find(".dataTables_empty").attr("colspan", summary.visible.length);
       this._renderGrandTotalsRow();
+      this._renderHeaderGroup();
```
nburwell: please add a test to cover this case (esp since it tripped us up).

ie confirm your header, then programmatically set visible(false) on one of the columns, and check that the DOM has hidden that whole header and sub columns.

src/plugins/data_table/data_table.js
codyjca: Done.

src/plugins/data_table/data_table.js
codyjca: Done.

spec/plugins/data_table/local.js
codyjca: Done.

codyjca:
@nburwell addressed your comments, up for another round of CR?

dist/backdraft.js
```diff
@@ -1644,9 +1651,15 @@ _.extend(Plugin.factory, {
           var headerGroupDataIndex = columnConfig.headerGroupDataIndex;
           var columnGroupConfig = _.findWhere(columnGroups, { "headerGroupDataIndex" : headerGroupDataIndex } );
 
+          if (columnGroupConfig === undefined || headerGroupDataIndex === undefined) {
```
nburwell: in this case any null/false/undefined should go into this error block right? I'd do:

```
if (!columnGroupConfig || !headerGroupDataIndex) {
```

codyjca: Done.

PR #71, opened 2017-02-01
WEB-2442 auto scroll table when dragging a column to reorder it and near viewport edge

nburwell:
@garvank here is my PR for auto scroll of table when moving columns

I am going to see if I can clean up the logic as I put in support for both types of tables (overflow auto on the wrapper), versus a more traditional table that spills out past the viewport.  Supporting both made it get a bit messier, plus is coupled right now to a wrapper class we use by convention which I don't like.

### DIFF
--- a/examples_server/examples/data_table/large_table_column_reordering/data_table.js
+++ b/examples_server/examples/data_table/large_table_column_reordering/data_table.js
@@ -1,0 +1,125 @@
+Backdraft.app("TableExample", {
+
+  plugins : [ "DataTable"],
+
+  activate : function($el) {
+    this.mainRouter = new this.Routers.Main({ $el : $el });
+    Backbone.history.start({ });
+  }
+
+});
+
+Backdraft.app("TableExample", function(app) {
+
+  app.router("Main", {
+
+    routes : {
+      "" : "index"
+    },
+
+    index : function() {
+      var view = new app.Views.Index();
+      this.swap(view);
+    }
+
+  });
+
+});
+
+Backdraft.app("TableExample", function(app) {
+
+  app.model("Book", {
+
+  });
+
+});
+
+Backdraft.app("TableExample", function(app) {
+
+  app.collection("Books", {
+
+    model : app.Models.Book,
+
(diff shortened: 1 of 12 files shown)


spec/plugins/data_table/local.js
```diff
@@ -294,10 +294,12 @@ describe("DataTable Plugin", function() {
         // we need to test this using an async strategy because the checkbox is toggled async as well
         table.dataTable.on("page", function() {
           _.defer(function() {
-            table.page("previous");
             expect(table.$("th.bulk :checkbox").prop("checked")).toEqual(true);
+            done();
```
nburwell: this test was not correct before, and was throwing warnings in `grunt spec`, now it's fixed

nburwell:
@garvank I removed the coupling to our wrapper class and got end to end tests for both overflow auto mode, and traditional browser scrollbar mode

examples_server/examples/data_table/large_table_column_reordering/data_table.js
```diff
@@ -0,0 +1,125 @@
+Backdraft.app("TableExample", {
+
+  plugins : [ "DataTable"],
```
garvank: Styleguide calls for no spaces between key name and `:`, looks like all keys in this file have a single space.

spec/plugins/data_table/general.js
```diff
@@ -814,6 +814,177 @@ describe("DataTable Plugin", function() {
       });
     });
 
+    describe("autoscroll on dragging of columns", function() {
+      beforeEach(function() {
+        app.view.dataTable.row("R", {
+          columns : [
```
garvank: More spaces after key names here

examples_server/examples/data_table/large_table_column_reordering/data_table.js
nburwell: Good point, I copied this from the other examples.. which were all made back before the styleguide.  I'll see at making a pass across all the examples in backdraft!

PR #72, opened 2017-02-21
1702/m/story 4952 new summary reports grand total row

alecjacobs5401:
@nburwell 

Mind giving this a review? This is the backdraft datatables upgrade portion to account for uniques.

Thanks!

### DIFF
--- a/src/plugins/data_table/data_table.js
+++ b/src/plugins/data_table/data_table.js
@@ -285,7 +285,7 @@ var LocalDataTable = (function() {
                 $grandTotalsRow.append($("<td>"));
               } else {
                 hasGrandTotalsCell = true;
-                var grandTotalsCellText = this.totalsRow.hasUniques ? "Total Gross*<br>Total Net" : "Grand Total";
+                var grandTotalsCellText = this.totalsRow.hasUniques ? "Total Gross<br>Total Net" : "Grand Total";
                 $grandTotalsRow.append($("<td>").addClass('.grand-total-title').html(grandTotalsCellText));
               }
             } else {
--- a/src/plugins/data_table/data_table.js
+++ b/src/plugins/data_table/data_table.js
@@ -100,26 +100,6 @@ var LocalDataTable = (function() {
       return this.dataTable.fnSettings().fnRecordsTotal();
     },
 
-    hasUniqueValues: function() {
-      var hasUniques = false;
-
-      if (this.totalsRow && this.totalsRow.model) {
-        var totals = this.totalsRow.model.attributes;
-        var columns = this.totalsRow.model.keys();
-
-        _.each(columns, function(column) {
-          if (columns.indexOf(column + ".unique") !== -1) {
-            if (!(totals[column] instanceof Array) && totals[column] != totals[column + '.unique']) {
-              hasUniques = true;
-              return false;
-            }
-          }
-        });
-      }
-
-      return hasUniques;
-    },
-
     columnRequired: function(state, id) {
       if (!state && this._columnManager.columnConfigForId(id).required) {
         throw new Error("can not disable visibility when column is required");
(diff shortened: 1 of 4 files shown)


src/plugins/data_table/server_side_data_table.js
```diff
@@ -188,6 +188,7 @@ var ServerSideDataTable = (function() {
           if (json.sEcho * 1 < oSettings.iDraw) return;
           if (json.total) {
             self.totalsRow = new self.rowClass({ model: new Backbone.Model(json.total)});
+            self.totalsRow.hasUniques = self.hasUniqueValues();
```
nburwell: the fact that you have to add this in afterwords is a code smell...

src/plugins/data_table/data_table.js
```diff
@@ -100,6 +100,26 @@ var LocalDataTable = (function() {
...
+
+        _.each(columns, function(column) {
+          if (columns.indexOf(column + ".unique") !== -1) {
+            if (!(totals[column] instanceof Array) && totals[column] != totals[column + '.unique']) {
+              hasUniques = true;
+              return false;
+            }
+          }
+        });
+      }
```
nburwell: this method only acts on the `totalRows`.. so it should live down there (then you don't need the line later in this PR - `self.totalsRow.hasUniques = self.hasUniqueValues();` . 

Fortunately we have a base class that all Row classes derive from: https://github.com/Invoca/backdraft/blob/042fd8ea3f855a2aa8047b1310e53acdf518edaa/src/plugins/data_table/row.js#L9-L9

src/plugins/data_table/data_table.js
```diff
@@ -262,14 +282,15 @@ var LocalDataTable = (function() {
             if (this.isNontotalsColumn && this.isNontotalsColumn(col)) {
               // If column is a non totals column, draw "Grand totals" on the first one and the rest are empty
               if (hasGrandTotalsCell) {
-                $grandTotalsRow.append($("<td></td>"));
+                $grandTotalsRow.append($("<td>"));
               } else {
                 hasGrandTotalsCell = true;
-                $grandTotalsRow.append($("<td>Grand Total</td>"));
+                var grandTotalsCellText = this.totalsRow.hasUniques ? "Total Gross<br>Total Net" : "Grand Total";
```
nburwell: not a huge fan this logic is in backdraft at all... probably fine for now but better would be to pass in a grand total label renderer (it could ask about uniques and then add the `br` etc).  just a thought

src/plugins/data_table/data_table.js
alecjacobs5401: Totally agree. It was something I contemplated doing at first, but not sure why I went against it.

alecjacobs5401:
@nburwell ready for the next round!

src/plugins/data_table/data_table.js
```diff
@@ -279,18 +259,19 @@ var LocalDataTable = (function() {
...
             if (this.isNontotalsColumn && this.isNontotalsColumn(col)) {
-              // If column is a non totals column, draw "Grand totals" on the first one and the rest are empty
               if (hasGrandTotalsCell) {
-                $grandTotalsRow.append($("<td>"));
+                $grandTotalsRow.append(node);
               } else {
                 hasGrandTotalsCell = true;
-                var grandTotalsCellText = this.totalsRow.hasUniques ? "Total Gross<br>Total Net" : "Grand Total";
-                $grandTotalsRow.append($("<td>").addClass('.grand-total-title').html(grandTotalsCellText));
+                node.addClass('.grand-total-title');
```
nburwell: minor, but do you still need this class added here now that you have a grandTotalRenderer that is invoked? ie couldn't the web app's grandTotalRenderer add this class?  slash I don't see it used in web at all?

alecjacobs5401: Yeah I can remove it. I was hoping to use it in web side, but ended up not. Thanks for catching this

ibiziiac-sv:
👍  )

PR #73, opened 2017-02-27
1702/m/STORY-4767.STORY-4860_save-button - Add small helper to get th…

codyjca:
…e pageLimit.

@nburwell or @alecjacobs5401 may I get some CR?

### DIFF
--- a/src/plugins/data_table/data_table.js
+++ b/src/plugins/data_table/data_table.js
@@ -100,6 +100,11 @@ var LocalDataTable = (function() {
       return this.dataTable.fnSettings().fnRecordsTotal();
     },
 
+    getPageLimit: function() {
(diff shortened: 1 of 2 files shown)


src/plugins/data_table/data_table.js
```diff
@@ -100,6 +100,11 @@ var LocalDataTable = (function() {
       return this.dataTable.fnSettings().fnRecordsTotal();
     },
 
+    getPageLimit: function() {
+      this._lockManager.ensureUnlocked("bulk");
+      return this.dataTable.fnSettings()._iDisplayLength;
+    },
```
nburwell: this looks like the only accessor using `get...` in the name.  The others are just the noun, like `totalRecordsCount` and `columnVisibility` etc.

What about `pageLimit`?

PR #74, opened 2017-03-09
1702/m/web 2299 fixed table header

garvank:
@nburwell here is the draft where if you scroll too far left the sizes bug out.
You can see my best attempt in `handleMouseMove` to fix the behavior but it isn't quite perfect.

I have more context on the relevant mousemove functions in `ColReorder` that I think might be involved if needed.

### DIFF
--- a/src/plugins/data_table/dataTables.colReorder.js
+++ b/src/plugins/data_table/dataTables.colReorder.js
@@ -992,33 +992,15 @@
                top: navbarHeight,
                width: tableWrapper.width(),
                overflow: 'hidden',
-               'border-bottom': '1px solid #eaeaea',
                left: tableLeftOffset,
                'z-index': 1
-             });
+             }).addClass('stuck');
 
              matchHeaderWidths();
              handleTableScroll();
 
              table.css({
                'margin-top': thead.height()
-             });
-
-             theadTh.each(function (i) {
-               var minWidth = parseInt($(this).css('min-width').replace('px', '')) + 5;
-               var newWidth = (headerWidths[i] > minWidth) ? headerWidths[i] : minWidth;
-
-               $(this).css({
-                 width: newWidth
-               });
-             });
-
-             tbodyTr.each(function (i) {
-               $(this).children('td').each(function(i) {
-                 $(this).css({
-                   width: headerWidths[i]
-                 });
-               });
              });
 
              stuck = true;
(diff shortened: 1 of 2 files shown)


src/plugins/data_table/dataTables.colReorder.js
```diff
@@ -920,6 +933,165 @@
...
+
+       // User must trigger 'tableLoaded' event after DataTable has been appended to DOM
+       // We need this to calculate offsets correctly
+       $(window).off('tableLoaded', start.bind(this));
+       $(window).on('tableLoaded', start.bind(this));
+
+       function start() {
+         var th, minWidth, minX;
+         var $window           = $(window);
+         var navbarHeight      = offsetTop;
```
nburwell: should name this more generically since it's in the library code (and actually why not keep `offsetTop` .. no reason to map to another var name)

src/plugins/data_table/dataTables.colReorder.js
```diff
@@ -920,6 +933,165 @@
...
+         var tableWrapper      = $(this.s.dt.nTableWrapper.children[0]);
+         var table             = $(this.s.dt.nTable);
+         var thead             = $(this.s.dt.nTHead);
+         var theadTop          = tableWrapper.offset().top;
+         var tbody             = $(this.s.dt.nTBody);
+         var tbodyTr           = tbody.children('tr');
+         var theadTh           = thead.children('tr').children('th');
+         var tableBorderOffset = tableWrapper.css('border-left') ? parseInt(tableWrapper.css('border-left').match('.*px')[0].replace('px', '')) : 0;
```
nburwell: this looks brittle (and hard to read).. is that going to work in all cases? can we at least make it it's own method and I'd feel more comfortable handling the case where match returns `null` (because in general it can..)

src/plugins/data_table/data_table.js
```diff
@@ -58,6 +58,7 @@ var LocalDataTable = (function() {
       this._enableRowHighlight();
       this.paginate && this._initPaginationHandling();
       this._triggerChangeSelection();
+      this._colReorder && this._colReorder._fnSetupStickyTableHeader.call(this._colReorder, this._colReorder.s.iStickyTableHeader);
```
nburwell: shouldn't this happen in colReorder? any chance there is already an init where you can just call `_fnSetupStickyTableHeader ` ? 

If not.. all of this is private stuff in there.. what about:

```
this._colReorder && this._colReorder.stickyHeader.call(this._colReorder)
```

and inside that "public" method it does the stuff above? That's the 2nd fav option though.

garvank: Yay, now that we're working on `tableLoaded` event, we can move it over into the `ColReorder` constructor if the config option is enabled.

jadametz:
Noticed this was in #dev-pull-reminders. Worth closing over a year later?

garvank:
@jadametz yep, closing. Thanks!

PR #78, opened 2017-10-14
Added babel to build task for transpiling from modern es to es5

garvank:
@nburwell simple change, care to review?


nburwell:
nice! 👍

garvank:
FYI, the build was failing because the Node version on Semaphore was way behind. 
I bumped the version to NodeJS `8.5.0` to build going forward.

PR #80, opened 2017-11-03
fire change event when modifying bulk select all/none checkbox

gabe-kent:


gabe-kent:
@nburwell Test added!

PR #81, opened 2017-11-29
WEB-3460 - Reporting Column Filter Popover:  fix FilterMenu enabled/disable

codyjca:


src/plugins/data_table/filter_view.js
```diff
@@ -119,16 +123,13 @@ var DataTableFilter = (function(options) {
     },
 
     disableFilter: function(errorMessage) {
+      this.errorMessage = errorMessage;
       if (!this.enabled) return;
```
nburwell: this line can just be removed now.. it's harmless to simply do L128 idempotently and not have an early return here

src/plugins/data_table/filter_view.js
```diff
@@ -12,19 +12,19 @@ var DataTableFilter = (function(options) {
...
-        <button class="btn btn-sm btn-filter" name="button" type="submit" title="">Apply</button>\
-        <button class="btn btn-primary btn-sm btn-clear pull-right" name="button" type="submit" title="">Clear</button>\
-      </div>\
+     <div class="filterMenu <%= filterMenuClass %>"> \
+       <% if (enabled) { %> \
+         <%= menuTemplate %> \
+         <button class="btn btn-sm btn-filter" name="button" type="submit" title="">Apply</button>\
+         <button class="btn btn-primary btn-sm btn-clear pull-right" name="button" type="submit" title="">Clear</button>\
+       <% } else { %> \
+         <span class="error-text"> \
```
nburwell: because there are tests that rely on this element and we don't want to couple tests (esp cross-repo tests) to CSS classes that are providing style... this should also have a `data-mount="error-message"` attribute on it. If it's not too late can you add that please?

PR #82, opened 2017-12-09
WEB-3548: Numeric Filter Menu lost track of previous filter value whe…

codyjca:
…n not setting a value and changing select option.

@garvank CR?

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "backdraft-app",
-  "version": "0.3.6",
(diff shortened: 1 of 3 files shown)


garvank:
LGTM! But just note that as discussed, depending on how these elements are queried more bugs will likely be found throughout the app... let's work this ticket next passion day to make a jQuery plugin to solve this issue: https://ringrevenue.atlassian.net/browse/TECH-2204

PR #83, opened 2018-01-18
1801/ux/tech 2081 bootstrap 4 migration.tech 1586 table action menu

nburwell:
@garvank here is the PR to remove popoverMenu on backdraft side - just going back to a normal bootstrap popover for the filter, which works well for column filters.

(My branch doesn't have the icons fixed, so ignore that part)
image


garvank:
Looks simple and easy enough to me 👍

nburwell:
> Why bring this back?

```
+          // popovers are smart enough to reposition when the last of the content is at the edge of screen
+          // but in a table that is positioned with fixed scrolling, it gets confused when content is
+          // at the edge of viewport with more of the table still off the screen
```

image

nburwell:
What we tested at your computer was a smaller table where we were at the edge of screen but also at the end of the table's content.

PR #85, opened 2018-01-23
Expose Listing's plugin classes List and Item 

dtognazzini:
This is necessary so we can use ES6 classes.

### DIFF
--- a/src/plugins/listing/index.js
+++ b/src/plugins/listing/index.js
@@ -3,6 +3,11 @@ Backdraft.plugin("Listing", function(plugin) {
   {%= inline("src/plugins/listing/list.js") %}
   {%= inline("src/plugins/listing/item.js") %}
 
+  plugin.exports({
+    List: List,
+    Item: Item
+  });
+
   plugin.initializer(function(app) {
 
     app.view.listing = function(name, properties) {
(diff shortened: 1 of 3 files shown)


src/plugins/listing/index.js
```diff
@@ -3,6 +3,11 @@ Backdraft.plugin("Listing", function(plugin) {
   {%= inline("src/plugins/listing/list.js") %}
   {%= inline("src/plugins/listing/item.js") %}
 
+  plugin.exports({
+    List : List,
+    Item : Item
+  });
```
garvank: I know elsewhere in the app we have some floating `:`s but our styleguide now suggests we keep them left aligned 🙃 just a heads up for future work!

dtognazzini: Oops. I copied and pasted from another exports call. I'll fix.

nburwell:
Thanks for doing this, excited to move towards es6 classes everywhere!

In this PR is it worth also doing the dataTable plugin as we have `app.view.dataTable("MyTableView", { ... })` view classes too

dtognazzini:
@nburwell I think other places should be updated as needed. I'd like to have a in-person conversation about next steps.

PR #86, opened 2018-02-03
Bundle JS files using Webpack

dtognazzini:
These changes usher in using Webpack for bundling. The end goal is to allow consumers to import Backdraft source files directly. For now, Webpack is used to build the distribution and for transpiling a la babel-loader.

Source files were modified only to the extent necessary to get Webpack working.

The following issues were filed for follow up work:
#87 - Use Browserlist?
#88 - Declare Dependencies 
#89 - Hook up Hound for ESLint. 
#90 - Revamp source to allow consumers to pull in files directly.
#91 - Overhaul specs to use ES6 modules.


dtognazzini:
Closed in favor of #92

Issue #91, opened 2018-02-05
Overhaul specs to use ES6 modules.


dtognazzini:

dtognazzini:
Addressed by #93

Issue #90, opened 2018-02-05
Revamp source to allow consumers to pull in files directly.


dtognazzini:
Consumers should be able to `import` ES6 modules from backdraft instead of using the namespacing methods.

Continue to provide the namespacing methods for down-rev consumers.

dtognazzini:
Addressed by #100

Issue #89, opened 2018-02-05
Hook up Hound for ESLint.


dtognazzini:

dtognazzini:
Addressed by #98

Issue #88, opened 2018-02-05
Declare dependencies


dtognazzini:
Currently, no dependencies are declared for the backdraft-app node module. Dependent libraries have been copied into `vendor/`. 

The task here is to declare dependencies in `package.json`. Dependencies only necessary for tests should be declared as `devDependencies`.

dtognazzini:
Addressed by #102

PR #92, opened 2018-02-05
Bundle JS files using Webpack

dtognazzini:
These changes usher in using Webpack for bundling. The end goal is to allow consumers to import Backdraft source files directly. For now, Webpack is used to build the distribution and for transpiling a la babel-loader.

This PR partially addresses #84.

Source files were modified only to the extent necessary to get Webpack working.

The following issues were filed for follow up work:
#87 - Use Browserlist?
#88 - Declare Dependencies 
#89 - Hook up Hound for ESLint. 
#90 - Revamp source to allow consumers to pull in files directly.
#91 - Overhaul specs to use ES6 modules.


src/plugins/data_table/dataTables.colReorder.js
```diff
@@ -25,6 +25,10 @@
  * Minor bug fixes by Jeremy Hubble @jeremyhubble
  */
 
+import jQuery from "jquery";
+import {Coordinates, DomVisibility} from "../../utils/dom_visibility";
```
garvank: Spaces inside `{}`'s 🙃

src/plugins/data_table/data_table.js
```diff
@@ -1,6 +1,21 @@
...
+
+import ColumnManager from "./column_manager";
+import Plugin from "../../plugin";
+
+import SelectionManager from "./selection_manager";
+import LockManager from "./lock_manager";
+
+import DataTableFilter from "./filter_view";
+
+import {extractColumnCSSClass, toColumnCSSClass} from "../../utils/css";
```
garvank: Same comment re:`{}`

dtognazzini:
@garvank Thanks for the review!

> One question: Do you think it would be valuable to set src/utils as a modules directory? Could be nice to just import the Util you need without referencing it directly.

I'm not sure what you mean. Let's talk in person.

Regarding the style comments, my inclination is to forego making those changes for now and instead leave them to be addressed by a new task that hooks up JSLint for this repo. We will want to use JSLint in `web/` at some point. We can leverage learnings from using it here when we use it in web.

package.json
```diff
@@ -16,18 +15,26 @@
...
-    "babel-preset-es2016": "^6.24.1",
-    "babel-preset-minify": "^0.2.0",
-    "babel-plugin-transform-es2015-modules-systemjs": "^6.24.1"
+    "underscore": "1.8.3",
+    "webpack": "^3.10.0"
+  },
+  "peerDependencies": {
+    "backbone": "1.1.2",
+    "jquery": "^1.10.2",
+    "underscore": "1.8.3"
```
nburwell: nice!

src/backdraft.js
```diff
@@ -1,30 +0,0 @@
...
-  {%= inline("src/plugin.js") %}
-  Backdraft.plugin = Plugin.factory;
-
-  {%= inline("src/plugins/base/base.js") %}
-  {%= inline("src/plugins/data_table/index.js") %}
-  {%= inline("src/plugins/listing/index.js") %}
-
-  window.Backdraft = Backdraft;
-
-})(jQuery);
```
nburwell: so glad to see this gone!

dtognazzini:
@nburwell 
> are the environment setup instructions still correct or should it change to yarn install etc?

No they are not. I'll add another commit for that. Thanks!

nburwell:
> No they are not. I'll add another commit for that.
Sweet thanks

👍

nburwell:
👍 on last changes

PR #93, opened 2018-02-24
Allow ES6 in specs

dtognazzini:
**Summary:**
1. Use [Karma](https://karma-runner.github.io) to run all specs.
1. Use sourcemap via Webpack's `devtool: 'inline-source-map'`.
1. Rework development tasks to use `yarn` (i.e. `npm`) scripts instead of `grunt`. Grunt is still used internally for invoking Webpack and for setting up the examples server. See README & package.json for available tasks.
1. In development, Babel is configured to _not_ minify. This aids debugging in the browser.

**Note**
While ES6 can now be used in the spec files, the specs files were **not** converted en masse to use ES6 aside from the bare minimum update made across every spec file to import the `Backdraft` namespace.

All specs continue to test against the `Backdraft` namespace. While, ideally, specs would import only the code they are testing, making those changes isn't strictly necessary for allowing ES6 in specs.

Specs can be converted to import what they're testing in the future, on an as-needed basis. If we ever decide to remove support for the namespace, that would be a natural point for converting the remaining spec files.


nburwell:
@dtognazzini how do you feel changing this PR to merge back to `1802/m/TECH-2313_use_webpack` ?

https://github.com/Invoca/backdraft/compare/1802/m/TECH-2313_use_webpack...1802/m/TECH-2354_convert_specs_to_es6

It at least makes the review much easier.. don't see much downside in combining the branches and merging / together either since this one won't really require separate QA..

dtognazzini:
@nburwell 
Sounds good!

dtognazzini:
> Would love to kill grunt in this repo at this point but looks like we'd need to port the example server with the build/watch capability.. and I guess the overall grunt build task which also could just be a node script.

Totally agree with this direction!

PR #98, opened 2018-04-13
1804/m/story 2340 code style checker

rahulk1000:


rahulk1000:
@caitlinrhd @dtognazzini @alecjacobs5401 code review please

package.json
```diff
@@ -23,6 +24,14 @@
     "babel-preset-env": "^1.6.1",
     "babel-preset-minify": "^0.3.0",
     "backbone": "1.1.2",
+    "eslint": "^4.19.1",
+    "style-guide": "Invoca/style-guide#1804/m/STORY-2340_code_style_checker",
```
alecjacobs5401: This should probably be updated to point to just the repo name

src/plugins/data_table/filter_view.js
```diff
@@ -387,15 +385,15 @@ var DataTableFilter = (function(options) {
...
-        case 'string':
-          filterMenu = new StringFilterMenu({column: options.column, parentView: this});
-          break;
-        case 'numeric':
-          filterMenu = new NumericFilterMenu({column: options.column, parentView: this});
-          break;
-        case 'list':
-          filterMenu = new ListFilterMenu({column: options.column, parentView: this});
-          break;
+      case 'string':
```
alecjacobs5401: our style guide doesn't explicitly state that cases should be indented once inside a switch statement, but the only `good` example that is used includes it.

src/utils/dom_visibility.js
```diff
@@ -26,15 +26,15 @@ const DomVisibility = Class.extend({
 
 const Coordinates = {
   absolutePointAtViewportEdge: function(edge, x, buffer) {
-    switch(edge) {
-      case 'left':
-        return x - $(window).scrollLeft() <= buffer;
+    switch (edge) {
+    case 'left':
```
alecjacobs5401: same comment

examples_server/examples/data_table/server_side/data_table.js
```diff
@@ -1,9 +1,9 @@
 Backdraft.app("TableExample", {
 
-  plugins : [ "DataTable"],
+  plugins: [ "DataTable"],
```
alecjacobs5401: is there a linter we can add to check for this? Looks like it missed an unecessary leading space  in the array

package.json
```diff
@@ -7,9 +7,10 @@
     "dist/backdraft.js"
   ],
   "scripts": {
+    "eslint": "eslint-git-changes --jsx",
     "setup": "npm install -g grunt-cli",
-    "specs": "karma start --single-run",
-    "dev": "karma start",
+    "specs": "yarn run eslint; karma start --single-run",
```
alecjacobs5401: should we instead add this to the karma config so it is always ran on `karma start`? Would DRY this out and seems to be our intended workflow

dtognazzini: I think the `specs` script will run on CI. Since we're using Hound, should we be running eslint on CI?

package.json
```diff
@@ -7,9 +7,10 @@
...
+    "dev": "yarn run eslint && karma start",
```
dtognazzini: How come we use `&&` here but `;` in the `specs` command?

package.json
```diff
@@ -23,6 +24,14 @@
...
     "babel-preset-minify": "^0.3.0",
     "backbone": "1.1.2",
+    "eslint": "^4.19.1",
+    "style-guide": "Invoca/style-guide#1804/m/STORY-2340_code_style_checker",
+    "eslint-config-standard": "^11.0.0",
+    "eslint-git-changes": "^2.0.1",
+    "eslint-plugin-import": "^2.11.0",
+    "eslint-plugin-node": "^6.0.1",
+    "eslint-plugin-promise": "^3.7.0",
+    "eslint-plugin-standard": "^3.0.1",
```
dtognazzini: Are all these dependencies necessary?

spec/plugins/data_table/server_side.js
```diff
@@ -35,12 +35,13 @@ describe("DataTable Plugin", function() {
...
     return {
-      status : 200,
-      responseText : JSON.stringify({
+      status: 200,
+      responseText: JSON.stringify({
         sEcho: this.echo++,
         iTotalRecords: 100,
         iTotalDisplayRecords: 100,
         aaData: [
+          /* eslint-disable camelcase */
```
dtognazzini: These are going to be super annoying at some point. I wonder if there's a way we can disable this rule for property names on objects.

src/plugins/data_table/dataTables.colReorder.js
```diff
@@ -30,1901 +30,1877 @@ import {Coordinates, DomVisibility} from "../../utils/dom_visibility";
...
-        thead.bind('mousemove.ColReorder', function (e) {
-          var nTable = that.s.dt.nTable;
-          if (that.dom.drag === null && that.dom.resize === null) {
-            /* Store information about the mouse position */
-            var nThTarget = e.target.nodeName === "TH" ? e.target : $(e.target).parents('TH')[0];
-            var offset = $(nThTarget).offset();
-            var nLength = $(nThTarget).innerWidth();
-
-            /* are we on the col border (if so, resize col) */
-            if (Math.abs(e.pageX - Math.round(of
```
dtognazzini: Could improve readability by using a descriptively-named local variable.

src/plugins/data_table/filter_view.js
rahulk1000: I think we should move towards not indenting cases inside of a switch (given that's what we do for ruby). I'm going to update our good example so that there's no indentation

examples_server/examples/data_table/server_side/data_table.js
rahulk1000: yep there is a linter for this: `"standard/array-bracket-even-spacing": ["error", "either"],`. We would have seen it if hound ran on this PR...

package.json
rahulk1000: yep

package.json
rahulk1000: I tried hooking eslint into our karma workflow but no luck. 
When there is a eslint error, it will continue on and run the tests (this is why we used a ; instead of a &&)

Ideally, don't think we want eslint on CI but given our current setup it will run on there. I suppose we can solve that by having a separate script for our CI...does that sound good?

package.json
rahulk1000: This should be a `;` as well...I will update it

spec/plugins/data_table/server_side.js
rahulk1000: There is a way to disable it for properties: https://eslint.org/docs/rules/camelcase

I'm thinking we should leave it on for properties for now and see how it goes. I had to use a comment in some cases here because those properties mapped to url parameters...

src/plugins/data_table/dataTables.colReorder.js
rahulk1000: I'm going to leave it as is for now

package.json
rahulk1000: correct, i'll make the change

rahulk1000:
@alecjacobs5401 @dtognazzini I have replied to the CR comments....

examples_server/examples/data_table/server_side/data_table.js
dtognazzini: Why isn't Hound running on this PR? :-)

package.json
dtognazzini: 👍

package.json
dtognazzini: K

spec/plugins/data_table/server_side.js
dtognazzini: We use underscores in property names for model data in Backdraft/Backbone apps. We will get a Hound bark every time that code is updated.

The [`{ "properties": "never" }`](https://eslint.org/docs/rules/camelcase#never) rule disables this warning for properties of objects but keeps the warnings for all other uses of underscores-in-names.

I think that's exactly what we want.

Using `{ "properties": "never" }` you wouldn't have needed to add `eslint-disable camelcase` comments to most (all?) of the places that this comment was added in this PR.

src/plugins/data_table/dataTables.colReorder.js
dtognazzini: 👍

src/plugins/data_table/selection_manager.js
```diff
@@ -4,20 +4,20 @@ import Class from "../../utils/class";
...
     return this._count;
   },
 
-  models : function() {
+  models: function() {
     return _.values(this._cidMap);
   },
 
-  process : function(model, state) {
+  process: function(model, state) {
```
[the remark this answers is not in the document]
dtognazzini: As discussed in IM: we should  set `space-before-function-paren` to “never”.

src/plugins/data_table/selection_manager.js
```diff
@@ -29,12 +29,12 @@ var SelectionManager = Class.extend({
       if (existing) {
         // purge existing entry
         delete this._cidMap[model.cid];
-        this._count = Math.max(0, this._count -1);
+        this._count = Math.max(0, this._count - 1);
```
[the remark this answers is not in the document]
dtognazzini: Can we address this one? I think we only want to warn about extra semicolons in cases like:

```javascript
var x = 5;;

function foo() {
    // code
};
```

src/plugins/data_table/filter_view.js
```diff
@@ -308,9 +307,9 @@ var DataTableFilter = (function(options) {
         // if there are url params for this filter...
         if (matches[0]) {
           // go through each of those list values
-          matches[0].value.forEach( function(element, index, array) {
+          matches[0].value.forEach(function(element, index, array) {
             // check it
-            this.$el.find('input[value="'+element+'"]').prop("checked", true);
+            this.$el.find('input[value="' + element + '"]').prop("checked", true);
```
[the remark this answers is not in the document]
dtognazzini: I think we should address
> Strings must use singlequote quotes

...as part of this work.

dtognazzini:
> https://github.com/Invoca/backdraft/pull/98#issuecomment-381747162

Thanks! I've left some more comments.

examples_server/examples/data_table/server_side/data_table.js
rahulk1000: It looks like it finally ran yesterday after I pushed up another commit..

package.json
rahulk1000: script for CI added: https://github.com/Invoca/backdraft/pull/98/commits/59149ada0bad86df1752b361038d620133e672ad

spec/plugins/data_table/server_side.js
rahulk1000: alright sounds good, we can use { "properties": "never" }

src/plugins/data_table/selection_manager.js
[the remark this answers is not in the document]
rahulk1000: Should be even bother warning about extra semi-colons? I would be ok if we did not. 
In the meanwhile, I am going to set `semi: "always"` which should take care of comments like these

src/plugins/data_table/selection_manager.js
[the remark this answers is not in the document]
rahulk1000: ack

src/plugins/data_table/filter_view.js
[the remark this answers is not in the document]
rahulk1000: i have turned off the single quote requirement

rahulk1000:
@dtognazzini replied_to/addressed your comments. eslint config changes can be found here: https://github.com/Invoca/style-guide/pull/7

package.json
```diff
@@ -7,9 +7,11 @@
     "dist/backdraft.js"
   ],
   "scripts": {
+    "eslint": "eslint-git-changes --jsx",
     "setup": "npm install -g grunt-cli",
-    "specs": "karma start --single-run",
-    "dev": "karma start",
+    "specs": "yarn run eslint; karma start --single-run",
```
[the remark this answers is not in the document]
rahulk1000: ack thanks

PR #102, opened 2018-06-28
1806/m/tech 2377 fix up backdraft dependencies

dtognazzini:
Summary:
- use dependencies from source instead of copies.
- package modified version 1.9.4 of jquery.dataTables.
- use and test against bootstrap 4
- remove dist. All consumers must now consume from source.

Addresses #88


dtognazzini:
@nburwell @garvank I've updated the README with ES6 usage. Please have a look and LMK what you think.

garvank:
Looks great! 👍

nburwell:
👍

PR #105, opened 2018-09-10
Web 4149 add url pagination flag

scottpchow23:
Adding a flag to data tables to allow for toggling of url pagination. This means you can now have a data table that paginates but doesn't necessarily use url pagination to remember it's history.

---
This change is [Reviewable](https://reviewable.io/reviews/invoca/backdraft/105)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "backdraft-app",
-  "version": "1.2.0",
+  "version": "1.2.1-beta.1",
   "description": "Plugin-based UI application framework built on top of and extending Backbone to create single page apps",
   "files": [
     "src/",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "backdraft-app",
-  "version": "1.2.1-beta.3",
+  "version": "1.2.1",
   "description": "Plugin-based UI application framework built on top of and extending Backbone to create single page apps",
   "files": [
     "src/",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "backdraft-app",
-  "version": "1.1.0-beta.1",
+  "version": "1.2.0",
   "description": "Plugin-based UI application framework built on top of and extending Backbone to create single page apps",
   "files": [
     "src/",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "backdraft-app",
-  "version": "1.2.1-beta.1",
+  "version": "1.2.1-beta.2",
   "description": "Plugin-based UI application framework built on top of and extending Backbone to create single page apps",
   "files": [
     "src/",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "backdraft-app",
-  "version": "1.2.1-beta.2",
+  "version": "1.2.1-beta.3",
   "description": "Plugin-based UI application framework built on top of and extending Backbone to create single page apps",
   "files": [
     "src/",
--- a/src/data_table/local_data_table.js
+++ b/src/data_table/local_data_table.js
@@ -489,7 +489,7 @@ class LocalDataTable extends View {
   }
 
   _afterRender() {
-    if (this.paginate) {
+    if (this.paginate && this.urlPagination) {
       this._goToPageFromQueryString();
     }
   }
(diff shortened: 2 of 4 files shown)


src/data_table/local_data_table.js
```diff
@@ -77,7 +82,7 @@ class LocalDataTable extends View {
     this._enableRowHighlight();
     this.paginate && this._initPaginationHandling();
     this._triggerChangeSelection();
-    this.paginate && this._setupPaginationHistory();
+    this.urlPagination && this._setupPaginationHistory();
```
nburwell: This should check both flags right?  Or, the above lines should ensure that urlPagination is ONLY set true if `paginate` is also true.. if the above does that, code below can be simpler probably.. otherwise you will need to always be sure to check both

src/data_table/local_data_table.js
```diff
@@ -29,6 +29,11 @@ class LocalDataTable extends View {
     _.extend(this, _.pick(this.options, ["selectedIds", "paginate"]));
     _.bindAll(this, "_onRowCreated", "_onBulkHeaderClick", "_onBulkRowClick", "_bulkCheckboxAdjust", "_onDraw",
       "_onColumnVisibilityChange", "_onColumnReorder");
+    if (this.options.urlPagination !== undefined) {
+      this.urlPagination = this.options.urlPagination;
+    } else if (this.urlPagination === undefined) {
+      this.urlPagination = true;
```
nburwell: per other comment I think this would be cleaner and less bug-prone if the else case was:

```
this.urlPagination = this.pagination;
```

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "backdraft-app",
-  "version": "1.1.0-beta.1",
+  "version": "1.2.1-beta.2",
```
scottpchow23: So I made a mistake when I was publishing versions and put 1.2.0 out (when it wasn't perfect), so the version I'm shooting to release at the end of this ticket is 1.2.1.

src/data_table/local_data_table.js
```diff
@@ -537,8 +542,11 @@ class LocalDataTable extends View {
   }
 
   _dataTableConfig() {
-    let displayStart = this._getSafeDisplayStartFromPageNumber();
-    let recordTotal = displayStart + this.paginateLength;
+    let displayStart, recordTotal;
+    if (this.urlPagination) {
+      displayStart = this._getSafeDisplayStartFromPageNumber();
+      recordTotal = displayStart + this.paginateLength;
+    }
```
nburwell: is it okay to pass `undefined` to iRecordsTotal and iRecordsDisplay below in the event urlPagination is off? (Versus omitting those params entirely). I assume so but wanted to check since before you were always sending 0 + length

src/data_table/local_data_table.js
```diff
@@ -484,7 +489,7 @@ class LocalDataTable extends View {
   }
 
   _afterRender() {
-    if (this.paginate) {
+    if (this.paginate && this.urlPagination) {
```
nburwell: can just check urlPagination if code above is preventing a mismatch of flags..

src/data_table/local_data_table.js
```diff
@@ -29,6 +29,11 @@ class LocalDataTable extends View {
     _.extend(this, _.pick(this.options, ["selectedIds", "paginate"]));
     _.bindAll(this, "_onRowCreated", "_onBulkHeaderClick", "_onBulkRowClick", "_bulkCheckboxAdjust", "_onDraw",
       "_onColumnVisibilityChange", "_onColumnReorder");
+    if (this.options.urlPagination !== undefined) {
+      this.urlPagination = this.options.urlPagination;
```
nburwell: per other comment, this would be ` = this.pagination && this.options.urlPagination;`

src/data_table/local_data_table.js
nburwell: Looks like you just pushed a commit for this.  I think still better to protect against these flags drifting though

scottpchow23: I just addressed that the moment before you made this comment 😂 , but yes it did seem inconsistent that I would drop the paginate check.

src/data_table/local_data_table.js
scottpchow23: If we pass undefined, I was thinking that it doesn't get mapped on the data table side of things, hence it's back to being a normal data table. That might be a wrong interpretation however.

package.json
dtognazzini: 👍

PR #106, opened 2019-01-08
TECH-2580 convert aoc app to webpacker

gabe-kent:


spec/router_spec.js
```diff
@@ -142,6 +143,68 @@ describe("Router", function() {
     });
   });
 
+  describe("navigate", function() {
+    class MyRouter extends Router {
+    }
+
+    class MyView extends View {
```
dtognazzini: Not sure this class is necessary. Can you get by by replacing uses with `View`?

spec/router_spec.js
```diff
@@ -142,6 +143,68 @@ describe("Router", function() {
...
+        this.$el.html('');
+        return this;
+      }
+    }
+
+    beforeEach(function() {
+      this.router = new MyRouter({ $el: $('<div>') });
+    });
+
+    it("should call router.navigate right away if view does not define a beforeNavigate function", function() {
```
dtognazzini: Using "should" is a bit of fluff. The same meaning could be communicated using: `it("calls ...`.

dtognazzini: The "if view does not define a beforeNavigate function" fragment describes a context and is perfect for folding into a `describe` block. 

```javascript
describe("navigate", function() {
  describe("with activeView, without beforeNavigate", function() {
    it("calls router.navigate right away", ...
  });
});

Comment (dtognazzini) [reply spec/router_spec.js]: Having written that comment ^, it's clear we're missing tests testing a "without active view" context.

File: spec/router_spec.js [hunk 3]
```diff
@@ -142,6 +143,68 @@ describe("Router", function() {
...
+      spyOn(Backbone.history, 'navigate');
```
dtognazzini: The implementation calls `super.navigate`. Spying on something known to be part of the implementation of `super.navigate` is brittle because the implementation of `super.navigate` could change, breaking these tests.

Can you change this to use a spy on `super.navigate`?

spec/router_spec.js
```diff
@@ -142,6 +143,98 @@ describe("Router", function() {
...
+      const view = new View();
+      view.render();
+      this.router.activeView = view;
+
+      this.router.navigate("nextPage");
+      expect(Backbone.Router.prototype.navigate).toHaveBeenCalledWith("nextPage", true);
+    });
+
+    describe("without beforeNavigate hook", function() {
+      describe("when activeView is provided", function() {
```
dtognazzini: I think the nesting here should be switched. The `beforeNavigate` hook doesn't mean anything without an `activeView`.

The nesting should be something like:
```
with active view:
  without beforeNavigate hook
  with beforeNavigate hook

without active view:
```

dtognazzini:
LGTM!

PR #108, opened 2019-03-08
WEB-4461: unescape filter value

echistyakov:
JIRA: https://ringrevenue.atlassian.net/browse/WEB-4461

Our backend HTML-encodes (unfortunately) stuff like column names, and filter options. If these values are used to make API requests without being decoded, the request will fail because an encoded value won't match a raw value in the DB. For example: encode signal name "Duration &gt; 30 sec" won't match the actual signal name "Duration > 30 sec".
Solution is to decode values before using them in requests. For our ListFilterMenu class, HTML-decode values before using them to construct the element.

`htmlDecode` implementation taken from here: https://github.com/Invoca/web/blob/0b0c45317881be4a730bf6960dbb1221543c0a27/client/reporting/views/helpers.js#L226-L232

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "backdraft-app",
-  "version": "1.3.1",
+  "version": "1.3.2-beta.0",
   "description": "Plugin-based UI application framework built on top of and extending Backbone to create single page apps",
   "files": [
     "src/",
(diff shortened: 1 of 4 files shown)


echistyakov:
@dtognazzini My original fix was wrong. The root cause of this problem is not the Javascript code, but backend code which always HTML-encodes certain values (as opposed to relying on JS to do encoding/escaping as appropriate). Unfortunately, changing backend code is not feasible because it might have negative effect around HTML escaping throughout the app. I added escaping to ListFilterMenu.

dtognazzini:
@echistyakov As we discussed, the UI should be sending back whatever it was handed. We discussed you're going to look int updating the server code to unescape as necessary. I'm going to remove myself as reviewer from this PR. Feel free to add me to a new PR if need be.  Thanks!

PR #109, opened 2019-04-01
WEB-4461: escape list filter menu

echistyakov:
JIRA: https://ringrevenue.atlassian.net/browse/WEB-4461

jQuery `value` selector cannot contain some [special characters](http://api.jquery.com/category/selectors/). In order to work they need to be escaped with a backslash.
Example issue is having a signal name "\" - `ListFilterMenu` won't check the option with that signal.

### DIFF
--- a/src/data_table/filter_menus/list_filter_menu.js
+++ b/src/data_table/filter_menus/list_filter_menu.js
@@ -14,7 +14,9 @@ class ListFilterMenu extends BaseFilterMenu {
         // go through each of those list values
         matches[0].value.forEach((element, index, array) => {
           // check it
-          this.$el.find(`input[value="${element}"]`).prop("checked", true);
+          // TODO: use jQuery.escapeSelector() once upgraded to jQuery >=3.0
+          const selector = `${element}`.replace(/[!"#$%&'()*+,./:;<=>?@[\\\]^`{|}~]/g, "\\$&");
+          this.$el.find(`input[value="${selector}"]`).prop("checked", true);
         });
         // make the button show
         this.parentView._toggleIcon(true);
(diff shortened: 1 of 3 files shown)


src/data_table/filter_menus/list_filter_menu.js
```diff
@@ -14,7 +14,9 @@ class ListFilterMenu extends BaseFilterMenu {
         // go through each of those list values
         matches[0].value.forEach((element, index, array) => {
           // check it
-          this.$el.find(`input[value="${element}"]`).prop("checked", true);
+          // TODO: use jQuery.escapeSelector() once upgraded to jQuery >=3.0
+          const selector = `${element}`.replace(/[!"#$%&'()*+,./:;<=>?@[\\\]^`{|}~]/g, "\\$&");
+          this.$el.find(`input[value="${selector}"]`).prop("checked", true);
```
dtognazzini: This code escapes the selector so it can be used in `find()`. Is there anything escaping the `value` of the `input` that the selector is looking for? If so, can we use that same escaping here? If not, then I'm not sure how this works.

echistyakov: `value` is HTML escaped in the backend (no XSS vulnerability). However, HTML-special-characters are different from CSS-special-characters used in jQuery selectors.
The following valid HTML element: `<input value="\my\signal"/>` can only be correctly found by escaping the backslash: `$el.find('input[value="\\\\my\\\\signal"]')`

dtognazzini: Ok. Can you please add tests for this change?

echistyakov: Done.

dtognazzini: Thanks

PR #126, opened 2021-10-15
Non production migrate build to GitHub actions

jebentier:
In the interest of moving forward with the Buildkite migration, Hydra has gone ahead and taken care of the migration for this helper gem. It seem rarely touched, so this seemed fairly safe to go ahead and migrate.

*Note:* Because this is a public repo, we've moved the build to Github Actions instead of Buildkite due to the policy that public projects should not build on internal infrastructure


lavham (approved):
LGTM

